PR #1, opened 2016-08-02
Add models and get the tests running

adam-thomas:
Copied from an internal project with a lot of renaming.

@incuna/backend review please?

PR #2, opened 2016-08-03
Add admin functionality

adam-thomas:
Allow people to edit questions and their structure in the admin!  Copied-and-renamed from the same place as #1.

I've had to remove django 1.10 support for the time being because it causes a "maximum recursion depth exceeded" issue when trying to display a `SurveyFieldset` object.  I have no idea why (Google has yielded nothing).  Ideas are welcome...


adam-thomas:
Bump?

adam-thomas:
Thanks :)

PR #3, opened 2016-08-04
Add an AppConfig

adam-thomas:
Contains settings for `field_type` choices and some slots for future settings (to configure the stuff I'll be adding next).


surveys/apps.py
```diff
@@ -0,0 +1,54 @@
...
+            'display_text': 'Character field (free text)',
+            'form_field_factory': None,
+            'serializer_field_factory': None,
+        }),
+        ('number', {
+            'display_text': 'Number',
+            'form_field_factory': None,
+            'serializer_field_factory': None,
+        }),
+        ('percentage', {
```
wytrych: I'm not sure percentage is a great description. Maybe just slider?

adam-thomas: These are drawn from 'the spec' for one of the projects we're using this in.  It'll be validated as a percentage (is the plan) but the slider is largely a UI decision.  So if anything the slider part is the suspect bit :P

wytrych: Ok in that case maybe drop the "slider" bit?

adam-thomas: Sure, can do :)

adam-thomas:
Updated btw :)

adam-thomas:
bump?

surveys/apps.py
```diff
@@ -0,0 +1,54 @@
...
+
+class SurveyConfig(AppConfig):
+    name = 'surveys'
+
+    # Complete dictionary of customisable data per field type.
+    # Has to be Ordered so we can guarantee consistent behaviour on things making
+    # use of it (dropdown fields etc).
+    field_type_data = OrderedDict((
+        ('free_text', {
+            'display_text': 'Character field (free text)',
```
KevinEtienne: Should this string be translatable?

surveys/apps.py
```diff
@@ -0,0 +1,54 @@
...
+        # Choices suitable for the field_type dropdown on SurveyField.
+        self.field_type_choices = [(k, v['display_text']) for k, v in field_data_items]
+
+        # Convenient mappings of field types to form/serializer field factories.
+        self.form_field_factories = {
+            k: v['form_field_factory'] for k, v in field_data_items
+        }
+        self.serializer_field_factories = {
+            k: v['serializer_field_factory'] for k, v in field_data_items
+        }
```
KevinEtienne: Maybe these three list comprehension can be refactored to an helper function? As only the value for the label is changing.

adam-thomas: `field_type_choices` is actually slightly different :P

surveys/apps.py
adam-thomas: oh, probably, yes

adam-thomas:
Added translations :)

adam-thomas:
_nudges gently_

adam-thomas:
New day bump

surveys/apps.py
maxpeterson: It seem odd to be iterating over `field_data_items`  three times rather than once:

```
self. field_type_choices = {}
self. form_field_factories = {}
self. serializer_field_factories = {}
for k, v in field_data_items:
    self. field_type_choices[k] = v['display_text']
    self.form_field_factories[k] = v['form_field_factory']
    self. serializer_field_factories[k] = v['serializer_field_factory']
```

surveys/migrations/0002_change_percentage_display_text.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('surveys', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='surveyfield',
+            name='field_type',
+            field=models.CharField(choices=[('free_text', 'Character field (free text)'), ('number', 'Number'), ('percentage', 'Percentage'), ('radio', 'Choose one option'), ('checkbox', 'Choose one or more options')], default='free_text', max_length=255),
```
maxpeterson: Will this necessitate a migration when the library is used in a project that overwrite `survey_config.field_type_choices`. Can you import `survey_config` in the migration and set `choices= survey_config.field_type_choices`

surveys/apps.py
```diff
@@ -0,0 +1,55 @@
...
+class SurveyConfig(AppConfig):
+    name = 'surveys'
+
+    # Complete dictionary of customisable data per field type.
+    # Has to be Ordered so we can guarantee consistent behaviour on things making
+    # use of it (dropdown fields etc).
+    field_type_data = OrderedDict((
+        ('free_text', {
+            'display_text': _('Character field (free text)'),
+            'form_field_factory': None,
```
maxpeterson: Do we actually need two factories? Will we ever use both `form_field_factory` and `serializer_field_factory`? I guess I am thinking we have a different `SurveyConfig` for forms and serializers. And the this determines whether it generates serializer fields or form fields.

surveys/migrations/0002_change_percentage_display_text.py
KevinEtienne: Maybe the choices can be set in the form or the serializer and keep `field_type` as a simple `CharField`?

surveys/apps.py
adam-thomas: Ah, yes.  Too much style guide habit :P

surveys/apps.py
adam-thomas: That seems like it wouldn't add much.  It also makes using `default_app_config` awkward.

surveys/migrations/0002_change_percentage_display_text.py
adam-thomas: Updated the migration (for now) :)

surveys/apps.py
adam-thomas: Fix'd!

adam-thomas:
Any more comments?

KevinEtienne:
Added two issues to keep track of:
 - two or more separate app configs
 - having choices set on the serializer and the form instead of the model.

adam-thomas:
Thanks :)

PR #4, opened 2016-08-05
Add Django field generators.

adam-thomas:

### DIFF
--- a/surveys/generators.py
+++ b/surveys/generators.py
@@ -1,0 +1,27 @@
+from collections import ChainMap
+
+
+class BaseFieldGenerator(object):
+    """
+    A base class for generating dynamic fields.
+
+    Creates and returns an instance of a field specified in the self.field_class
+    attribute. By default, the created field is required.  Make sure to set the
+    field_class attribute when overriding, otherwise your generator will be a bit useless.
+
+    Methods:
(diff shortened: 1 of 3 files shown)


adam-thomas:
bump

adam-thomas:
Moar bump

surveys/generators.py
```diff
@@ -0,0 +1,27 @@
...
+      instance.  (The instance is intended to be a SurveyField, which will specify
+      the field's behaviour.)
+    - generate_field - return an actual field instance.  The instance is given
+      the kwargs from get_field_kwargs() but those can be overridden or added to by
+      passing more to the generate_field() call.
+    """
+    required = True
+
+    def get_field_kwargs(self, instance):
+        return {'required': self.required}
```
maxpeterson: Now this is a library whether a field is `required`  should come from the model.  Do you want to address this here or as part of another PR?

maxpeterson: @adam-thomas can you create an issue to make `is_required` a field on the model.

adam-thomas:
Thanks :)

surveys/generators.py
adam-thomas: On the `SurveyField` model you mean?

PR #10, opened 2016-08-09
Add karma and dummy test

wytrych:


wytrych:

README.md
```diff
@@ -12,3 +12,9 @@ All frontend is located under `web/`:
 * `grunt` will start a dev server running under `localhost:9000`.
 
 This project uses `babel` and `browserify`. Source files are located in `web/app/scripts/src`. Compiled files are put into `web/dist`, which is symlinked to `web/app/scripts/js-build` for easy testing in the browser.
+
+### Tests
+
+The are two `grunt karma` targets:
+* `grunt karma:dev` once run will continue to watch the compiled `.js` files directory and the test files.
+* `grunt karma:ci` target only runs once
```
henrahmagix: `and is used in Travis`?

web/Gruntfile.js
```diff
@@ -100,6 +101,9 @@ module.exports = function (grunt) {
         }
     });
 
+    const karmaTask = require('./grunt/karma-config')(grunt);
```
henrahmagix: `grunt.loadTasks('./grunt')` will load all of them

henrahmagix: Oh but that doesn't return the module, nevermind

web/grunt/karma-config.js
```diff
@@ -0,0 +1,41 @@
...
+                        }
+                    },
+                    karmaTasksConfig
+                )
+            }
+        });
+    };
+
+    return {
+        configure
```
henrahmagix: Why return a configure function? I think all the grunt config and task registration should happen immediately, like all other grunt plugins. Then you can `grunt.loadTasks('./grunt')` and write these like grunt plugins.

wytrych: That's the pattern we've been always using up to this point. You mean the config should happen automatically when loading the plugin? Not sure I like that.

web/grunt/karma-config.js
```diff
@@ -0,0 +1,41 @@
+module.exports = function (grunt) {
+    'use strict';
+
+    const unitTests =  grunt.config('config.tests');
+
+    const karmaTasksConfig = {
+        browsers: ['Firefox'],
+        frameworks: ['jasmine', 'browserify'],
+        preprocessors: {
+            [unitTests]: ['browserify'],
```
henrahmagix: Use object-inline grunt config processing instead of this, so this can match the line below. That means you can remove `unitTests = grunt.config('config.tests').

web/grunt/karma-config.js
henrahmagix: That's how grunt plugins are written.

When we need to build complex multi-running things, where we want to require a load of modules then run certain steps of them in sequence, like in some big projects with multiple frontend-ops parts, then this pattern works well. Otherwise, I think it's overcomplicating a simple grunt plugin.

wytrych:
@henrahmagix Updated

wytrych:
@incuna/js bump

wytrych:
@incuna/js @henrahmagix  bump

.travis.yml
```diff
@@ -22,3 +22,7 @@ install:
   - pip install -r requirements.txt --use-wheel
   - pip install $DJANGO --use-wheel
   - cd web && npm install && cd ..
+before_script:
+  - export DISPLAY=:99.0
```
henrahmagix: I would've thought you'd need to wrap this in quotes so the yaml parser doesn't read `:` as a map

henrahmagix: But it works so 👍  =)

web/grunt/karma-config.js
```diff
@@ -0,0 +1,37 @@
...
+                            singleRun: true
+                        }
+                    },
+                    karmaTasksConfig
+                )
+            }
+        });
+    };
+
+    configure();
```
henrahmagix: Why make a function then call it? It's simpler without the `configure()` function.

web/Gruntfile.js
henrahmagix: Now that `karma-config` is a grunt plugin, you can `grunt.loadTasks('./grunt')`, which means any number of grunt plugins can be added to that directory and they'll all be loaded.

henrahmagix: To explain it more, you can
```js
var localGruntPlugins = './grunt';
grunt.loadTasks(localGruntPlugins);
```

web/package.json
```diff
@@ -18,6 +18,12 @@
...
     "grunt-eslint": "19.0.0",
     "grunt-jscs": "^3.0.1",
-    "jit-grunt": "0.10.0"
+    "grunt-karma": "^2.0.0",
+    "jasmine": "^2.4.1",
+    "jit-grunt": "0.10.0",
+    "karma": "^1.1.2",
+    "karma-browserify": "^5.1.0",
+    "karma-firefox-launcher": "^1.0.0",
+    "karma-jasmine": "^1.0.2"
```
henrahmagix: Pin the dev dependencies?

web/grunt/karma-config.js
wytrych: Here I won't agree. For me it's a bit cleaner to define functions and then call it, than have some code floating around. If you wanted to add something more, before the configure step for example, it would be much nicer to add a function and then call it lile:
```
prepare();
configure();
```
than just add some more floating code. I think it helps with organisation.

henrahmagix: I would agree for functional aspects of this, say if it was a lot of code and you wanted to be more descriptive, or to be DRY, but not for `grunt.config.merge()` use.

wytrych: @grahamgilchrist Opinion please

wytrych:
@incuna/js @grahamgilchrist bump

web/grunt/karma-config.js
grahamgilchrist: haha. Normally I would agree to try to put everything in functions, but since the `configure()` literally only wraps grunt.config.merge()` its almost unnecessary. But I could see how it would be useful if `configure()` was going to do more things.

grahamgilchrist: In short, it doesn't worry me, so happy to merge as it is.

wytrych: I can see it's almost unnecessary, but I also see the temptation to potentially add some more floating code in the future. It may never happen but it kind of keeps things cleaner I think. @grahamgilchrist Can have merge?

henrahmagix:
👍

PR #12, opened 2016-08-10
Add serializer field generators

adam-thomas:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ coverage==4.0.3
 dj-inmemorystorage==1.4.0
 dj-database-url==0.4.0
 django-orderable==4.0.3
+djangorestframework==3.4.3
 factory-boy==2.6.0
 flake8==2.5.4
 flake8-import-order==0.5.3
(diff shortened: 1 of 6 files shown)


wytrych:
Are you using the api description from here? https://github.com/incuna/ipsen-optimise/pull/14 It would be really good to have this in this project actually. I want to write tests against it, would be great if the backend was using that too, so we don't fall out of sync

adam-thomas:
Not deliberately.  I can update the API description with whatever the serializer fields output.  I'm not sure the `legend` field is in there by default.

wytrych:
Well, if we don't use the api-description then we'll get into trouble sooner or later. I would really advocate for it, it has proven to work well previously.

adam-thomas:
How set in stone is it meant to be?

wytrych:
Well, more than less I think, but at least if we have the api-description in this project and we're both testing against it, then if you need to update it a little bit I will know I have to fix my code too. If this is meant to be a library then it should really be ready to use out of the box, and the api-description can come a long way to help with that.

adam-thomas:
I'll do it after this PR is merged.

wytrych:
Cool thanks

KevinEtienne:
Looks really cool :) We have a generator now for API fields.

adam-thomas:
Thanks @KevinEtienne :)

PR #13, opened 2016-08-10
Add API description

adam-thomas:

Decisions/assumptions I made:
- The form list serializer only provides forms and their URLs.  The user won't be displayed more than one form at a time, so sending the full data seems impractical.
- Form and formset names cannot be null (this comes from the model - they're `unique=True`, so `null=True` isn't practical).
- The list URL is `/forms` and the form detail and POST URLs are both `/forms/<pk>`.

Let me know your thoughts. :)


api-description/forms/pk/get.json
```diff
@@ -0,0 +1,54 @@
...
+                    }]
+                },
+                {
+                    "name": "Numeric fields",
+                    "fields": [{
+                        "id": 2,
+                        "name": "What is answer to the ultimate question of life, the universe, and everything?",
+                        "help_text": null,
+                        "field_type": "number"
+                    },
```
wytrych: I think these should be indented more to the right, as they are elements of a list

api-description/forms/pk/post.json
```diff
@@ -0,0 +1,23 @@
+{
+    "description": "Endpoint to answer a form",
+    "url": "http://localhost:8000/forms/1",
+    "fields": {
+        "answers": {
+            "1": "Friends",
+            "2": "one_time",
+            "3": ["two_times", "three_times"],
+            "4": 5
```
wytrych: Would this accept a decimal string too? Like `"5"` ?

adam-thomas: Yup, should do!

api-description/forms/pk/get.json
adam-thomas: They're already indented?  I'm not sure what you mean.

wytrych: Yes, but is should have double indentation since it is a dict in an object:
```
                   "fields": [
                         {
                                 "id": 2,
                                 "name": "What is answer to the ultimate question of life, the universe, and everything?",
                                 "help_text": null,
                                 "field_type": "number"
                           },
                           {
                                     .....
                           }
                 ]
```

adam-thomas: Ah, sure.  Updated!

wytrych:
I'm happy if @incuna/backend is

adam-thomas:
Thanks :)

api-description/forms/pk/get.json
```diff
@@ -0,0 +1,60 @@
...
+    "description": "A detail view for a form",
+    "url": "http://localhost:8000/forms/1",
+    "OK": {
+        "status": 200,
+        "response_data": {
+            "name": "How have you been using the site?",
+            "url": "http://localhost:8000/forms/1",
+            "fieldsets": [
+                {
+                    "name": "Free text field",
```
KevinEtienne: We need either `help_text` or `description` for each `fieldset`. It would help the user to display generic information about the group of fields if necessary.

see 	https://incuna-pm.atlassian.net/wiki/display/OP/Questionnaires

> 003	Each fieldset will optionally include a multiple line free text description.

KevinEtienne: We would need the same kind of field `help_text` or `description` for the form

> 	001	Each questionnaire will optionally include a multiple line free text description.

api-description/forms/pk/post.json
```diff
@@ -0,0 +1,23 @@
...
+        }
+    },
+    "OK": {
+        "status": 201,
+        "response_data": {
+            "answers": {
+                "1": "Friends",
+                "2": "one_time",
+                "3": ["two_times", "three_times"],
+                "4": 4
```
KevinEtienne: Mini typo here, it should be 5.

api-description/forms/pk/get.json
adam-thomas: Oh, uh, that's been updated since I started working on this.

api-description/forms/pk/post.json
adam-thomas: You might want to fix that in the project I copied and pasted from :)

adam-thomas:
Updated :)

api-description/forms/pk/post.json
adam-thomas: Actually it might be `"5"` in practice.  I think so.  I'm not sure, depends on the internals of the process.  Will fix it if it turns out to be so.

api-description/forms/pk/get.json
KevinEtienne: Oh I see we don't have a form model yet, let's add it later if needed.

api-description/forms/pk/post.json
adam-thomas: Just realised the return value of this endpoint might show this value as `"5"`.  I genuinely don't know yet, it depends on the internals of the process.  I'll change the description later if so.

api-description/forms/pk/get.json
adam-thomas: Also that ^

api-description/forms/pk/post.json
wytrych: You mean I submit `4` but get back `5` ?

adam-thomas: no, it's 5 in the request description :P

wytrych: But if I post `4` I should get back `4` right? (either quoted or not, but the value is surely the same?)

adam-thomas:
Thanks :)

PR #14, opened 2016-08-11
Add Survey model and tweak/clean up the rest

adam-thomas:
* Add the `Survey` model itself (which m2ms to `Fieldset` via an orderable through table).
* `Fieldset` is no longer `Orderable`.
* `Field` and `Fieldset` admins are no longer `OrderableAdmins` either.
* `UserResponse` has a `ForeignKey` to `Survey` now.
* Added admin classes for `Survey`.


KevinEtienne:
Looks good, should we have a `CHANGELOG.md`?

adam-thomas:
ah, good call :)

adam-thomas:
thanks!

adam-thomas:
Actually to be fair there already is one, and this project has no releases yet, so there's nothing much to put in it until we publish a v0.1.

PR #15, opened 2016-08-11
Add angular

wytrych:


wytrych:

web/app/scripts/src/libraries.es.js
```diff
@@ -0,0 +1 @@
+export const angular = window.angular;
```
grahamgilchrist: I see this is a wrapper to expose angular as an ES6 module which is nice. It's a shame I had hoped angular itself would expose an ES6 module hook that you can point directly at, but it doesn't look like it :(

grahamgilchrist:
Looks good

web/app/scripts/src/libraries.es.js
wytrych: It actually does in the `npm` package. I thought of using it this way, however there are some compatibility issues:
* If you require a module it will be browserified, probably there is a way to tell browserify to ignore some files, but then in a package that will use this library es6 will be required to browserify it too. Also other packages might need updating to work in the same way. It would be probably quite painful to get it working.

grahamgilchrist: Hmm yes. Thats sounds like it could eventually be a very nice way to do it if we browserify everything and link with npm packages, but you are right it would be a big change from where we are now

PR #16, opened 2016-08-11
Add serializers and update API description

adam-thomas:

### DIFF
--- a/api-description/forms/pk/get.json
+++ b/api-description/forms/pk/get.json
@@ -5,7 +5,7 @@
         "status": 200,
         "response_data": {
             "name": "How have you been using the site?",
-            "url": "http://localhost:8000/forms/1",
+            "description": "Some additional text expanding on the above",
             "fieldsets": [
                 {
                     "name": "Free text field",
@@ -15,7 +15,8 @@
                             "id": 1,
                             "name": "How did you discover the site?",
                             "help_text": "Search engine, friend...",
-                            "field_type": "free_text"
+                            "field_type": "free_text",
+                            "answers": []
                         }
                     ]
                 },
@@ -27,13 +28,15 @@
                             "id": 2,
                             "name": "What is answer to the ultimate question of life, the universe, and everything?",
                             "help_text": null,
-                            "field_type": "number"
+                            "field_type": "number",
+                            "answers": []
                         },
                         {
                             "id": 3,
                             "name": "How much percentage does it take?",
                             "help_text": null,
-                            "field_type": "percentage"
+                            "field_type": "percentage",
+                            "answers": []
                         }
                     ]
                 },
@@ -46,14 +49,14 @@
                             "name": "How many time do you visit the site per day?",
                             "help_text": "Choose a valid answer.",
                             "field_type": "radio",
-                            "choices": ["One time", "Two times", "Three times or more"]
+                            "answers": ["One time", "Two times", "Three times or more"]
                         },
                         {
                             "id": 5,
                             "name": "Choose more than one answer.",
                             "help_text": null,
                             "field_type": "checkbox",
-                            "choices": ["One time", "Two times", "Three times or more"]
+                            "answers": ["One time", "Two times", "Three times or more"]
                         }
                     ]
                 }
--- a/api-description/forms/pk/get.json
+++ b/api-description/forms/pk/get.json
@@ -8,54 +8,62 @@
             "description": "Some additional text expanding on the above",
             "fieldsets": [
                 {
+                    "pk": 1,
                     "name": "Free text field",
                     "description": "Some additional text expanding on the above",
                     "fields": [
                         {
-                            "id": 1,
+                            "pk": 1,
                             "name": "How did you discover the site?",
                             "help_text": "Search engine, friend...",
                             "field_type": "free_text",
+                            "required": true,
                             "answers": []
                         }
                     ]
                 },
                 {
+                    "pk": 2,
                     "name": "Numeric fields",
                     "description": "Some additional text expanding on the above",
                     "fields": [
                         {
-                            "id": 2,
+                            "pk": 2,
                             "name": "What is answer to the ultimate question of life, the universe, and everything?",
                             "help_text": null,
                             "field_type": "number",
+                            "required": false,
                             "answers": []
                         },
                         {
-                            "id": 3,
+                            "pk": 3,
                             "name": "How much percentage does it take?",
                             "help_text": null,
                             "field_type": "percentage",
+                            "required": true,
                             "answers": []
                         }
                     ]
                 },
                 {
+                    "pk": 3,
                     "name": "Choice fields",
                     "description": "Some additional text expanding on the above",
                     "fields": [
                         {
-                            "id": 4,
+                            "pk": 4,
                             "name": "How many time do you visit the site per day?",
                             "help_text": "Choose a valid answer.",
                             "field_type": "radio",
+                            "required": true,
                             "answers": ["One time", "Two times", "Three times or more"]
                         },
                         {
-                            "id": 5,
+                            "pk": 5,
                             "name": "Choose more than one answer.",
                             "help_text": null,
                             "field_type": "checkbox",
+                            "required": false,
                             "answers": ["One time", "Two times", "Three times or more"]
                         }
                     ]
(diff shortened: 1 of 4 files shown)


surveys/serializers.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    class Meta:
+        model = models.UserResponse
+        fields = ['fieldset', 'user_id', 'answers']
+
+
+class SurveyResponseSerializer(serializers.Serializer):
+    """Serializer for a user's answers to a whole survey."""
+    user_responses = UserResponseSerializer(many=True)
+    survey = serializers.PrimaryKeyRelatedField(queryset=models.Survey.objects.all())
```
KevinEtienne: If we want to remove survey, we can use instead a `ModelSerializer` as we have the same information in the url, eg. `http://localhost:8000/forms/1`

api-description/forms/pk/post.json
```diff
@@ -2,22 +2,36 @@
     "description": "Endpoint to answer a form",
     "url": "http://localhost:8000/forms/1",
     "fields": {
-        "answers": {
-            "1": "Friends",
-            "2": "one_time",
-            "3": ["two_times", "three_times"],
-            "4": 5
-        }
+        "survey": 1,
```
wytrych: If I need to `post` the survey number back, I would need to `GET` it too

wytrych: It's a bit strange that I have to do that though, because it's the `pk` right? So if I'm sending it to url with a `pk` why do I also have to send the `pk` ?

api-description/forms/pk/post.json
```diff
@@ -2,22 +2,36 @@
...
+        "user_responses": [
+            {
+                "fieldset": 1,
+                "user_id": "User#20#",
```
wytrych: Where is that coming from?

api-description/forms/pk/post.json
```diff
@@ -2,22 +2,36 @@
...
-        "answers": {
-            "1": "Friends",
-            "2": "one_time",
-            "3": ["two_times", "three_times"],
-            "4": 5
-        }
+        "survey": 1,
+        "user_responses": [
+            {
+                "fieldset": 1,
```
wytrych: Similar issue here, if I have to send the `fieldset` number, I need to `get` it too

api-description/forms/pk/post.json
adam-thomas: Depends on the project using the library.

api-description/forms/pk/post.json
adam-thomas: Yeah, Kevin said that yesterday.  I'm not 100% sure on how that works programmatically, I might have to 'force it' in the view.

api-description/forms/pk/post.json
wytrych: So we're not basing this on `csrf` tokens? Also It needs to be resend per `fieldset`?

api-description/forms/pk/post.json
wytrych: The above comment is now hidden, but it's still important

api-description/forms/pk/get.json
```diff
@@ -5,55 +5,66 @@
...
                     ]
                 },
                 {
+                    "pk": 3,
                     "name": "Choice fields",
                     "description": "Some additional text expanding on the above",
                     "fields": [
                         {
-                            "id": 4,
+                            "pk": 4,
```
wytrych: Is it better to call it `pk` than `id`?

api-description/forms/pk/post.json
```diff
@@ -2,22 +2,35 @@
...
     "OK": {
         "status": 201,
         "response_data": {
-            "answers": {
-                "1": "Friends",
-                "2": "one_time",
-                "3": ["two_times", "three_times"],
-                "4": 5
-            }
+            "survey": 1,
```
wytrych: Do you still want to keep this one? It's not very useful since you removed the need to `POST` it

api-description/forms/pk/get.json
```diff
@@ -5,55 +5,66 @@
         "status": 200,
         "response_data": {
             "name": "How have you been using the site?",
-            "url": "http://localhost:8000/forms/1",
```
wytrych: Hmm you removed the `url` from this form, but that would be actually very useful for me

adam-thomas: It's the URL of the page itself.

api-description/forms/pk/post.json
adam-thomas: It's just what the serializer will give you, I think.

api-description/forms/pk/get.json
adam-thomas: `pk` is what we normally call it conventionally on the backend, is there a reason to call it `id` instead?

wytrych: I will be translating it as `id` on the field itself, also I thought that `pk` is more about an object in the database, while this is a field on an object? Or am I wrong? Not a big deal though

api-description/forms/pk/get.json
wytrych: But also where to `POST` right? I think it makes sense to have, but if it's a big pain then I can do some mork work on the fe

api-description/forms/pk/get.json
adam-thomas: I'll make it `id` then.  IIRC Django models have both, they're just aliases of one another by default.

api-description/forms/pk/get.json
adam-thomas: Why does it require extra work?  I'm a bit confused.

wytrych: Again, not a massive deal, but for one I think we usually have `url`s for things we post to?

The extra work bit is that I get the url of of the page from the forms list and then if it was on the form itself I would have the url automatically. If not then I have to store the url I'm getting somewhere and use that to post. I think it kind of breaks the concept of a REST API, but as I said, not a biggie.

adam-thomas: Yeah, ok.  I should be able to make it a `HyperlinkedModelSerializer` when we have views/urls in place, so I'll readd the url to the description.

api-description/forms/pk/get.json
wytrych: Cool thanks, I've seen you changed it but in the code but not in the json

wytrych:
Cool I don't have any more comments

adam-thomas:
Thanks :)

adam-thomas:
Anyone else happy/unhappy?

wytrych:
Just this comment left https://github.com/incuna/incuna-surveys/pull/16#discussion_r74568010

adam-thomas:
Oh, you said you were done :P

wytrych:
Heh, I added that comment, and THEN said I don't have any more :)

adam-thomas:
The notification masked the previous one, lol

adam-thomas:
Can I merge?

api-description/forms/pk/post.json
```diff
@@ -2,22 +2,35 @@
...
-        "answers": {
-            "1": "Friends",
-            "2": "one_time",
-            "3": ["two_times", "three_times"],
-            "4": 5
-        }
+        "user_responses": [
+            {
+                "fieldset": 1,
+                "user_id": "User#20#",
```
KevinEtienne: We might want to get the user from the request, but this can be addressed later maybe?

KevinEtienne:
Except https://github.com/incuna/incuna-surveys/pull/16/files#r74585071 looks good.

api-description/forms/pk/post.json
adam-thomas: There isn't necessarily a user model, as in the project this is from.

KevinEtienne: What we can do is to remove it from the post request and save the user in the view, something like what we did in [`humira`](https://github.com/incuna/abbvie-humira-epatient/blob/5318d57eb2ad5c5409960f7f5a91001d4198f144/api/tracker/mixins.py#L31).

```python
    def perform_create(self, serializer):
        serializer.save(patient=self.get_patient())
```

KevinEtienne: After discussion IRL let's see how do we end up using users.

adam-thomas: :+1:

adam-thomas:
Thanks :)

PR #17, opened 2016-08-15
Add a survey form GET API view

adam-thomas:
Returns the full details of the survey, including the fieldset nested structure, as per the API description.

### DIFF
--- a/surveys/models.py
+++ b/surveys/models.py
@@ -3,7 +3,7 @@ from django.contrib.postgres.fields import ArrayField, JSONField
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from orderable.models import Orderable
-
+from rest_framework.reverse import reverse as drf_reverse
 
 survey_config = apps.get_app_config('surveys')
 
@@ -80,6 +80,9 @@ class Survey(models.Model):
 
     def __str__(self):
         return self.name
+
+    def get_api_url(self):
+        return drf_reverse('survey-form', kwargs={'pk': self.pk})
 
 
 class SurveyFieldOrdering(Orderable):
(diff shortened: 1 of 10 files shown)


wytrych:
Are you actually using the JSON in testing?

adam-thomas:
Not yet, no.

surveys/models.py
```diff
@@ -81,6 +81,9 @@ class Survey(models.Model):
     def __str__(self):
         return self.name
 
+    def get_api_url(self):
+        return drf_reverse('survey-form', kwargs={'pk': self.pk})
```
KevinEtienne: Where do you need `get_api_url`?

surveys/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from . import views_api
+
+
+urlpatterns = [
+    url(r'^forms/(?P<pk>\d+)/$', views_api.SurveyView.as_view(), name='survey-form'),
```
KevinEtienne: I think `angular` queries the urls without a `/` at the end and I think in general we make it optional with a `?`:
```python
    url(r'^forms/(?P<pk>\d+)/?$', views_api.SurveyView.as_view(), name='survey-form'),
```

surveys/models.py
adam-thomas: We'll need it for the API description tests coming up soonish.  I figured I'd throw it in.

surveys/urls.py
adam-thomas: Will add the `?`, thanks :)

adam-thomas:
Updated :)

PR #18, opened 2016-08-15
Adding basic service for api communication

wytrych:


wytrych:

api-description/forms/get.json
```diff
@@ -10,7 +10,7 @@
             },
             {
                 "name": "Another form",
-                "url": "http::localhost:8000/forms/2"
+                "url": "http://localhost:8000/forms/2"
```
grahamgilchrist: `://`

grahamgilchrist: ah ignore me

grahamgilchrist:
Looks good. On a side note, I'm wondering if we should deprecate the `project_settings` module pattern and instead use properly documented providers to set config values?

It slightly bothers me relying on something outside of packages which is not a dependency or versioned in any way :)

wytrych:
Yeah I can change that and make a module instead

grahamgilchrist:
It's unrelated to this so happy to merge, I was just thinking since we are starting this as a relatively new project, it's a good opportunity to improve how we handle config in packages.

We can do it as another PR/issue though.

PR #19, opened 2016-08-15
Use a ProjectConfig module instead of PROJECT_SETTINGS

wytrych:
Relying on an external settings file is prone to be buggy


wytrych:

wytrych:
@incuna/js @incuna/frontend Bump

web/app/scripts/src/main.es.js
```diff
@@ -1,4 +1,8 @@
...
-// TODO: remove the following two lines
 import { angular } from 'libraries';
-/* eslint no-console: 0 */
+
+import Api from 'services/api';
+import ProjectConfig from 'providers/project-config';
+
+angular.module('incuna-surveys', [])
+    .provider('ProjectConfig', ProjectConfig)
+    .service('Api', Api);
```
henrahmagix: We usually name services as `ApiService`, like we do `FooCtrl` or `FooController`, but is no biggy

web/app/scripts/src/providers/project-config.es.js
```diff
@@ -0,0 +1,22 @@
+const ProjectConfig = function () {
+    let settings = {
+        apiRoot: 'localhost:8000'
+    };
+
+    return {
+        $get: function () {
+            return {
+                getApiRoot: () => settings.apiRoot
```
henrahmagix: I think it'd be simpler to `return settings`, so you don't have to add a get method for every new property. But I see how that means the data can be changed. I think simpler would be better here: less work to update for changes

web/app/scripts/src/providers/project-config.es.js
```diff
@@ -0,0 +1,22 @@
...
+            }
+        },
+        setApiRoot: (value) => {
+            settings.apiRoot = value;
+        }
+    };
+}
+
+export const provider = [
```
henrahmagix: I'm not sure of the pattern here. I think either the export should be renamed to explain that it's an Angular provider registration array-of-dependencies, or just `export const provider = ProjectConfig`

web/app/scripts/src/providers/project-config.es.js
wytrych: I think the pattern we used was to have a `getter` for each set of properties at least

web/app/scripts/src/providers/project-config.es.js
wytrych: Yeah I was thinking about it, should the dependencies be supplied in the `export` or when you register the thing in the main module. There's pros and cons for both, the angular dependency injector is not really suited to work well with ES6 modules

web/app/scripts/src/providers/project-config.es.js
henrahmagix: That's what I think is too much work to be worth it: every time you need to add a new property, say `powerLevel: 9001`, you have to add `getPowerLevel: () => settings.powerLevel`

henrahmagix: For a simple object of data, I think that's too much. If this was a service of methods that did meaningful things (like massage data or change stuff around), or there was a requirement to privatise `settings` so it cannot be accessed directly or changed, then 👍 , but I don't think that's the case

henrahmagix:
👍 👍 👍 👍 👍 👍 👍  for using providers

web/app/scripts/src/providers/project-config.es.js
henrahmagix: Ah, I see what you mean.

What about registering a new module in this file, and exporting that module's name?
```js
import ProjectConfigModule from 'project-config.js';

angular.module('incuna-surveys', [
    ProjectConfigModule,
    // ...
]);
```

Then `project-config` can deal with itself. We've done that before so that outside code doesn't have to manage dependencies of separate files. It may feel pointless to have one module per Angular _thing_, but it solves the dependencies-pattern problem and means each file can be read and understood independently.

henrahmagix:
```js
// project-config.js
export const module = angular.module('incuna-surveys.project-config', []);
module.provider('ProjectConfig', ProjectConfig);

export default module;
```

wytrych: Well but it doesn't work so well when you want to use that module in a different module, you rely on other modules anyway. So yeah it's basically the same pattern we've done before, but it's just an angular thing that you have to rely on actual strings

henrahmagix: After discussing in Slack, we can register modules and components in each file, then export the string identifiers so they can be used as dependencies in a DRY manner.

```js
// project-config.js

// Define imports and dependencies near each other.
import {SomeModuleName, SomeServiceName} from 'something.js';

const dependencies = [SomeModuleName];

// Define module code.
const ProjectConfig = function () {
    // ...
};

// Define identifier strings.
export const provider = 'ProjectConfig';

export const module = 'incuna-surveys.project-config';

// Register everything with Angular.
angular.module(module, dependencies)
    .provider(provider, ProjectConfig)
    .run([
        SomeServiceName,
        function (SomeService) {
            // Do something in the run phase with SomeService
        }
    ]);

// Export the module name by default.
export default module;
```

```js
// main.js
import ProjectConfigModule from 'project-config.js';

angular.module('incuna-surveys', [
    ProjectConfigModule,
    // ...
]);
```

wytrych:
@henrahmagix Updated

web/app/scripts/src/services/api.es.js
```diff
@@ -17,10 +22,16 @@ export const api = function ($http, PROJECT_SETTINGS) {
...
+export const module = {
+    moduleName: 'incuna.surveys-api',
+    componentName: 'API'
+};
+
+angular.module(module.moduleName, [ProjectConfig.moduleName])
+    .service([
+        '$http',
+        ProjectConfig.componentName,
+        API
```
henrahmagix: I think `API` should be defined inline here, as I don't think it should be exported (I see no reason to expose it), and the Angular inline-array injector syntax is difficult to read and work on when the function is separated from the dependencies.

web/tests/api-spec.js
```diff
@@ -4,13 +4,15 @@ describe('api service', function () {
         fixture.setBase('tests/api-description');
 
         inject(function ($http) {
-            const MOCK_PROJECT_SETTINGS = {
-                API_ROOT: 'localhost:8000'
+            const mockProjectConfig = {
+                getSettings:() => ({
+                    apiRoot: 'localhost:8000'
+                })
```
henrahmagix: I think this syntax is confusing. It's not immediately apparent that wrapping the object in parens stops the curly braces from being parsed as an arrow-function block, and that removing the parens will break it.

What about
```js
const settings = {
    apiRoot: 'blah'
};
const mockProjectConfig = {
    getSettings: () => settings
};
```

web/app/scripts/src/services/api.es.js
wytrych: I expose it for easier mocking, I think it's a bit easier to just pass in a mock module than to use the `injector()` syntax.

I'm not bothered too much with this syntax:
1) you have one indendation level less if you define it as a function
2) you still see the modules being imported where the function is defined

I would like to keep it this way

web/tests/api-spec.js
wytrych: Well this is just ES6 syntax, it's going to be used

web/app/scripts/src/services/api.es.js
henrahmagix: >I expose it for easier mocking

👍 

>I think it's a bit easier to just pass in a mock module than to use the `injector()` syntax.

Not sure what you mean by that. Could you explain in examples?

> 2) you still see the modules being imported where the function is defined

I mean the injector dependencies, `'$http'` and `ProjectConfig.componentName`. I think it's important that they're listed right next to the function definition that accepts them as parameters in order.
E.g.
```js
module.service('Foo', [
    'dep1',
    'dep2,
    function (dep1, dep2) {}
]);
```
Instead of
```js
var FooService = function (dep1, dep2) {};
// ...
// ...
// ...
module.service('Foo', [
    'dep1',
    'dep2,
    FooService
]);
```

wytrych: https://github.com/incuna/incuna-surveys/blob/2fb1fe7c07878ff74cdf92d11bf17ff294efeb54/web/tests/api-spec.js#L7-L14

wytrych: Can I use the `function` definition? I think it would be so much nicer to read to be able to define functions after they're being used:
```
angular.module(module.moduleName, [ProjectConfig.moduleName])
    .service([
        '$http',
        ProjectConfig.componentName,
        API
    ]);

export function API ($http, ProjectConfig) {
    return {
        getBaseUrl: function () {
            const endpoint = 'forms';
            const apiRoot = ProjectConfig.getSettings().apiRoot;
            return `${apiRoot}/${endpoint}`;
        },
        getList: function () {
            const url = this.getBaseUrl();

            return $http.get(url)
                .then((response) => response.data);
        },
        getForm: function (url) {
            return $http.get(url)
                .then((response) => response.data);
        }
    }
};
```

henrahmagix: I really don't think it's worth it.

However, to get the best out of both, why not use [ng-annotate](https://github.com/olov/ng-annotate) so we can have minified js whilst using the magic injector syntax?
```js
var FooFunction = function (dep1, dep2) {};
module.service('Foo', FooFunction);
```

henrahmagix: Regarding mocking, angular-mocks should really be used, so you don't have to change your app code just for tests.

wytrych: In general I think it is worth it look at this code:
```
function doSomething() {
    importSomeData();
    processTheData();
}

function importSomeData() {
    data = 'foo';
}

function processSomeData() {
     processedData = transform(data);
}

function transform (data) {
    return data * 2;
}
```
vs
```
var transform = function (data) {
    return data * 2;
}

var processSomeData = function() {
     processedData = transform(data);
}

var importSomeData = function () {
    data = 'foo';
}

var doSomething = function() {
    importSomeData();
    processTheData();
}
```

Which one is easier to follow?

henrahmagix: Sorry, I wasn't talking specifically about using function definitions over expressions. You can still use that in my example.

```js
function FooFunction (dep1, dep2) {};
module.service('Foo', FooFunction);
```

wytrych: `ng-annotate` won't work here because it's just taking the arguments passed in and uses those as strings, so I can't pass in the name of the service from an external module and expect that to work

henrahmagix: Oh yeah :<

wytrych: So I dunno what to do. I don't think it's a bad pattern, I've seen it on the web. It is different from what we usually do, but I'm ok with that.

henrahmagix: @grahamgilchrist @maxpeterson?

web/tests/api-spec.js
grahamgilchrist: In general I am against arrow functions for the simple reason of readability.

```
const x = function () {
 return {
   apiRoot: 'blah';
 }
};
```

is much more human readable than:

```
const x() => ({
  apiRoot: 'localhost:8000'
});
```

What advantage does the latter give? The only one I see is shorter code, which is not an issue when the final thing will be minified for production. The readability by other devs is much more important in this case and I think should take priority over shorter code.

I know this is syntax allowed by the spec, but I don't think we should use it for the same reason we try to avoid other language shortcuts. It makes it harder to human read for anyone not immediately familiar with the syntax. New developers, people coming from other languages or even people just not familiar with this bit of code will understand the former much more easily.

I can see a small argument for using them when you really need to bind `this` (although I'd have preferred they implement a `scopedFunction()` or even using `bind()` instead of the arrows), but otherwise I don't think we should use them generally.

wytrych: Hmm that's a strong position. I can see arrow funcitons playing nice in promises:

`.then((response) => request.data)`
Or in a `d3` context `.attr('x', (data) => data.height)`

web/app/scripts/src/services/api.es.js
```diff
@@ -17,10 +22,16 @@ export const api = function ($http, PROJECT_SETTINGS) {
...
-    '$http',
-    'PROJECT_SETTINGS',
-    api
-];
+export const module = {
+    moduleName: 'incuna.surveys-api',
+    componentName: 'API'
+};
+
+angular.module(module.moduleName, [ProjectConfig.moduleName])
```
grahamgilchrist: I can see what we're trying to do with all this importing module names, but it is making the code a lot more complex than possibly necessary?

It could be a lot simpler to just hardcode these? I'm not sure. This is sort of related to @henrahmagix comment below

web/tests/api-spec.js
henrahmagix: @grahamgilchrist I think arrow functions are great. The only problem here is using parens makes the syntax way more confusing. My example solves that and I think is the best of both

web/app/scripts/src/services/api.es.js
grahamgilchrist: I agree with @henrahmagix, it's much easier and less error-prone to see the dependency lists together like

```
[
    'dep1',
    'dep2',
    function (dep1, dep2) {}
]
```

Mixing the angular and ES6 imports dynamically like this makes it quite hard to read. How about defining `API` as the whole object that is passed to `angular.service`?

web/tests/api-spec.js
grahamgilchrist: even then in promises the example is less readable. The only advantage is shorter length of code by removing keywords, which IMO is much worse for readable and maintainable code.

wytrych: Well, I can see it less readable for someone who only did program `C` or `Java` before, but python doesn't have `function` keyword at all for example, and in `.NET` or `coffescript`, or `ruby` you do have arrows so for devs coming from these background it would be readable.

We're using angular which for anyone who has no experience in it is a tough thing to understand. So the argument that using arrows functions, a standard ES6 feature which will be all over the place makes things more difficult is not so valid imo.

web/app/scripts/src/services/api.es.js
henrahmagix: >It could be a lot simpler to just hardcode these?

What would you export otherwise?

There's no point exporting a controller function to be used in `myModule.controller('SomeCtrl', importedCtrlFunction)`, because that's a separation of code that is more confusing than helpful. So instead, each file registers its own `angular.module` and angular components (like we've done elsewhere), and then exports the module and component names so they can be depended on by other files.

If they're being exported (the only import/export setup that makes sense to me), we may as well use them.

web/app/scripts/src/services/api.es.js
henrahmagix: The only reason (I can see) that `API` exists is because it's required for mocking in tests. We should be using angular-mocks, which would make `export const API` redundant, so the definition can then be moved inline.

wytrych: Yes that's the only reason, but what's wrong with not using `angular-mocks` in this case?

henrahmagix: Your tests then need to be kept up-to-date with the dependencies required for this definition, and you need to inject the dependencies yourself. As soon as you add another dependency to this, in any position, the tests will fail.

It's way, way simpler to let angular do all that and use angular-mocks' `module()` function to mock it. Adding dependencies to angular components won't make the tests fail. If you want, you can DRY your mocks so updates to this definition then only require updates to tests code in one place rather than several.

web/tests/api-spec.js
grahamgilchrist: Still, writing the function out in long form is clearer. I don't think there can be any argument with that?

I'm not saying you can't understand these things once you are familiar with them, but it makes the barrier to entry higher. 
If you are not using the `this` binding, the advantage of arrows is shorter code. 
The disadvantage is being less readable.
If given this choice in a team setting where people are still learning and not everyone is a JS expert, IMO the clear choice is to choose the more readable way?

There are many things in our style guide where we recommend to write for readability rather than using code shortcuts. e.g. for the the same reason we try to avoid terniary operators instead of `if..then` because it is harder to understand.

web/app/scripts/src/services/api.es.js
grahamgilchrist: Yes I do think using angular mocks would be a better approach as well for the reasons @henrahmagix has outlined. Using the built in angular dependency injection is much easier to keep in sync.

wytrych:
@henrahmagix @grahamgilchrist Updated

web/app/scripts/src/services/api.es.js
```diff
@@ -1,37 +1,41 @@
 import { angular } from '../libraries';
 
-import ProjectConfig from '../providers/project-config';
+import ProjectConfigModule from '../providers/project-config';
```
henrahmagix: I think you can keep `ProjectConfig` as the identifier

web/app/scripts/src/providers/project-config.es.js
```diff
@@ -1,28 +1,29 @@
...
+module.provider(moduleProperties.componentName, [
+    function () {
+        let settings = {
+            apiRoot: 'localhost:8000'
+        };
+
+        return {
+            $get: function () {
+                return {
+                    getSettings: () => settings
```
henrahmagix: I think it'd be simpler to `return settings` in `$get`. I see how this allows us to add more to it, but I doubt there would be more to add.

web/app/scripts/src/services/api.es.js
wytrych: I thought it's going to be confusing with the actual `ProjectConfig` used in the service

henrahmagix: Yeah I see what you mean 👍

web/tests/api-spec.js
```diff
@@ -29,34 +19,37 @@ describe('api service', function () {
...
+            this.api.getForm(url).then((data) => {
                 expect(data.name).toBe('How have you been using the site?');
                 expect(data.fieldsets[0].name).toBe('Free text field');
-            
-                done();
             });
-            this.$httpBackend.flush();
+
             this.$rootScope.$digest();
+            this.$httpBackend.flush();
```
henrahmagix: Would it be useful for readability to have `this.resolvePromises()` that calls this, so that the end of the test describes why it needs to do that?
```js
beforeEach(function () {
    this.resolvePromises = function () {
        this.$rootScope.$digest();
        this.$httpBackend.flush();
    };
});

it('should get some data', function () {
    this.api.getSomething().then((data) => {
        expect(data).toBeData();
    });
    this.resolvePromises();
});
```

Or `resolvePromises` could be `endHttpSpec`

wytrych: I think that's a bit too much to be fair, too much abstraction for me

web/app/scripts/src/providers/project-config.es.js
```diff
@@ -0,0 +1,25 @@
+import { angular } from '../libraries';
+
+export const moduleProperties = {
+    moduleName: 'incuna.surveys-config',
```
henrahmagix: `incuna.surveys-config` ->
`incuna-surveys.config` ?

web/app/scripts/src/services/api.es.js
```diff
@@ -1,26 +1,41 @@
...
-        },
-        getForm: function (url) {
-            return $http.get(url)
-                .then((response) => response.data);
-        }
-    }
+import ProjectConfigModule from '../providers/project-config';
+
+export const moduleProperties = {
+    moduleName: 'incuna.surveys-api',
```
henrahmagix: `incuna.surveys-api` ->
`incuna-surveys.api` ?

henrahmagix:
👍

web/dist/incuna-forms.js
```diff
@@ -11,17 +11,77 @@ var angular = exports.angular = window.angular;
 
 var _libraries = require('./libraries.js');
```
grahamgilchrist: Why use `_` at the start of these vars? You don't seem to have any conflicting variable names?

wytrych:
Any more comments?

grahamgilchrist:
Hmm, interesting. Should we be trying to namespace all our modules under `incuna.`?

I'm happy to merge :)

henrahmagix:
👍

PR #20, opened 2016-08-15
Add API description test for SurveySerializer

adam-thomas:
The huge pile of setup is kind of awful and the error output is totally useless, but it works.  I think.  I tried several ways to produce decent error output and each one failed, but suggestions are welcome :P

FYI @wytrych I made some minor changes to the JSON.  `help_text` is now `''` when empty rather than `null`, since the serializer gives `''`.  I also selfishly shortened one of the questions so I didn't have to break it over multiple lines in the test setup code.

### DIFF
--- a/surveys/tests/test_views_api.py
+++ b/surveys/tests/test_views_api.py
@@ -20,5 +20,7 @@ class TestSurveyView(RequestTestCase):
         # that it's the same data here and hasn't been tampered with.  That way, if the
         # serializer is broken, only that test should fail, and this test will be
         # unconcerned by changes to the serializer format.
-        expected_data = SurveySerializer(instance=self.survey).data
+        expected_data = SurveySerializer(
+            instance=self.survey, context={'request': request}
+        ).data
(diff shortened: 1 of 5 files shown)


surveys/tests/test_views_api.py
```diff
@@ -20,5 +20,7 @@ def test_get(self):
         # that it's the same data here and hasn't been tampered with.  That way, if the
         # serializer is broken, only that test should fail, and this test will be
         # unconcerned by changes to the serializer format.
-        expected_data = SurveySerializer(instance=self.survey).data
+        expected_data = SurveySerializer(
+            instance=self.survey, context={'request': request}
```
KevinEtienne: Small nitpick about the style, we use one item per line.

KevinEtienne:
Looks good to me

surveys/tests/test_views_api.py
adam-thomas: oh whoops

adam-thomas:
Thanks :)

PR #22, opened 2016-08-17
Add a survey POST view

adam-thomas:
We have to serve up the POST serializer.  Contains much tweaking of the serializer and view.

### DIFF
--- a/surveys/tests/test_views_api.py
+++ b/surveys/tests/test_views_api.py
@@ -1,3 +1,5 @@
+from django.http import Http404
+
 from .utils import APIRequestTestCase, create_survey_data
 from .. import views_api
 from ..models import UserResponse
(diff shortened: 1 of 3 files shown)


surveys/tests/test_views_api.py
```diff
@@ -25,3 +28,41 @@ def test_get(self):
...
+        # Assert we return the same data but with the survey and user_id re-specified
+        # on the objects that need them.
+        data['survey'] = self.survey.pk
+        data['user_responses'][0]['user_id'] = data['user_id']
+        self.assertEqual(response.data, data)
+
+        # Assert the UserResponse object was created correctly.
+        user_response = UserResponse.objects.first()
+        self.assertEqual(user_response.user_id, data['user_id'])
+        self.assertEqual(user_response.answers, data['user_responses'][0]['answers'])
```
KevinEtienne: I would create 3 different tests here to target more the asserts.

KevinEtienne:
Looks good to me.

surveys/tests/test_views_api.py
adam-thomas: Three tests?  I can see a way to split it into two but I don't think there's much use in separating out the `user_response` asserts.

KevinEtienne: Two sounds good.

adam-thomas:
Updated :)

PR #23, opened 2016-08-18
Implement UserResponse history

adam-thomas:
UserResponses now have a timeline and a queryset method that can return the latest responses for each user on a given survey!

Thoughts on memory footprint: Going by some back-of-the-envelope estimation we're looking to handle up to 5x5x10000=250,000 entries (5 fieldsets in a survey, 5 responses per user over time, 10,000 participating HCPs in Europe) and each entry should be around a kilobyte or less.

After this is done, I plan on changing the survey choice fields so they key by numerical list indices rather than the answer strings themselves - currently the choices are `(('Yes', 'Yes'), ('No', 'No'))` whereas `((0, 'Yes'), (1, 'No'))` is probably much more efficient.  This will cut the memory usage down dramatically if we have a lot of choice fields, as well as being much faster to aggregate down the line.

@incuna/backend @maxpeterson review please? :)


surveys/models.py
```diff
@@ -114,6 +117,37 @@ def num_users(self):
...
+
+        # Build a dictionary of dictionaries:
+        #  {<user_id>: {<fieldset>: <answers>}}
+        # Use defaultdict so we don't have to check if <user_id> is already in the
+        # results before assigning values to the <fieldset> entry.
+        results = defaultdict(dict)
+        for entry in raw_data:
+            user_id = entry['user_id']
+            fieldset = entry['fieldset__pk']
+            results[user_id][fieldset] = entry['answers']
```
KevinEtienne: Should this be a helper method instead? Like this we have the queryset method which returns a queryset and a helper method which convert the data.

adam-thomas: Well, the queryset method would just return the survey-filtered and sorted list.  The data "conversion" part here is actually the important bit of the work.  (Check the docstring.)

adam-thomas:
Thanks :)

surveys/models.py
maxpeterson: How does this perform with many `UserResponse`s?

adam-thomas: @maxpeterson Ran this in the Python shell:

```
UserResponseFactory.create_batch(25000, survey=s)
a = time.time(); UserResponse.objects.latest_per_user(s); print(time.time() - a)
```

(The factory didn't like it when I asked it to make 250,000 instances.)

The 25,000 entry version (10% of the size) took 2.1404 seconds.  So we can reasonably expect the full EU dataset to take around 20 seconds to retrieve (for a decent-sized survey after the site has been in operation for a year or two) and the dataset for a single country to be a second or two.

adam-thomas: Also my computer has well-documented RAM issues and was able to handle this with no complaints.

PR #25, opened 2016-08-18
Angular formly

wytrych:


wytrych:

wytrych:
screenshot from 2016-08-18 13-21-01

web/Gruntfile.js
```diff
@@ -9,9 +9,13 @@ module.exports = function (grunt) {
         require('load-grunt-tasks')(grunt);
     } else {
         // Use jit-grunt to only load necessary tasks for each invocation of grunt.
-        require('jit-grunt')(grunt);
+        require('jit-grunt')(grunt, {
+            ngtemplates: 'grunt-angular-templates'
+        });
     }
 
+    const ngTemplatesPaths = require('./node_modules/grunt-incuna-plugins/helper-functions/ng-templates-paths')();
```
henrahmagix: If you set a `main` file in that library's `package.json`, you can `require('grunt-incuna-plugins')` and use that to access `ng-templates-paths`. (One can recursively require all files in a folder and export them as their file names.)

web/Gruntfile.js
```diff
@@ -40,8 +45,12 @@ module.exports = function (grunt) {
                     '<%= config.srcScriptsDir %>/**/*.es.js'
                 ],
                 tasks: [
-                    'compilejs'
+                    'build'
                 ]
+            },
+            ngtemplates: {
+                files: ['app/templates/**/*.html'],
+                tasks: ['build']
```
henrahmagix: The watcher should really only change what is necessary, so compiling js for src js changes, and running `ngtemplates` when html templates change.

wytrych: but this triggers a `template.js` compilation, which then needs to be compiled into `dist` files, because the `es` watcher only watches `es.js` source files.

web/app/scripts/src/services/fieldsets-parser.es.js
```diff
@@ -0,0 +1,92 @@
...
+export const moduleProperties = {
+    moduleName: 'incuna-surveys.field-parser',
+    componentName: 'FieldsParser'
+};
+
+const module = angular.module(moduleProperties.moduleName, [
+    FieldsConfigModule.moduleName
+]);
+
+module.service(moduleProperties.componentName, [
```
henrahmagix: If returning an object of methods, use `factory`. If you use `service`, you can assign each method to `this` instead of onto an object: less nesting and no `return`.

web/app/scripts/src/test.es.js
```diff
@@ -0,0 +1,83 @@
+const apiDesc = {
```
henrahmagix: Can this be in a separate file? And is it not the same as the current api description json file(s)?

web/app/scripts/src/test.es.js
```diff
@@ -0,0 +1,83 @@
...
+        'formlyConfig',
+        'FieldsParser',
+        function (
+            formlyConfig,
+            FieldsetParserService
+        ) {
+
+            this.fields = FieldsetParserService.parseFields(apiDesc);
+            this.model = FieldsetParserService.parseModel(apiDesc);
+        }]);
```
henrahmagix: `]);` should be on the next line

web/app/scripts/src/test.es.js
```diff
@@ -0,0 +1,83 @@
...
+                    'answers': ['One time', 'Two times', 'Three times or more']
+                }
+            ]
+        }
+    ]
+    /* jscs:enable disallowQuotedKeysInObjects */
+};
+
+angular.module('test-app', ['incuna-surveys'])
+    .controller('FormCtrl', [
```
henrahmagix: Use a `module` variable instead of chaining

web/app/scripts/src/test.es.js
```diff
@@ -0,0 +1,83 @@
...
+                    'required': false,
+                    'answers': ['One time', 'Two times', 'Three times or more']
+                }
+            ]
+        }
+    ]
+    /* jscs:enable disallowQuotedKeysInObjects */
+};
+
+angular.module('test-app', ['incuna-surveys'])
```
henrahmagix: Is this test app meant to be in the source?

web/app/templates/incuna-surveys/fields/checkbox.html
```diff
@@ -0,0 +1,6 @@
+<div drf-form-field="to.fieldOptions" field-id="options.key" class="checkbox">
+    <div ng-repeat="choice in to.choices" class="checkable checkbox">
+        <input class="checkbox-input" id="{{ options.key + '-' + $index }}" type="checkbox" checklist-model="model[to.fieldSetIndex].answers[options.key]" checklist-value="$index">
+        <label class="checkbox-label" for="{{ options.key + '-' + $index }}">{{ choice }}</label>
```
henrahmagix: How about
```
id="{{ options.key }}-{{ $index }}"
```

web/tests/fieldsets-parser-spec.js
```diff
@@ -0,0 +1,123 @@
+describe('FieldsetParserService', function () {
+
+    beforeEach(function () {
+        fixture.setBase('tests/api-description');
+        this.fieldset = fixture.load('forms/pk/get.json').OK.response_data;
+
+        angular.module('formly', []);
```
henrahmagix: Is this not a module from a library you should be loading, instead of defining yourself?

web/Gruntfile.js
henrahmagix: Surely `dist` files should only be changed when you want to make a build?

If you wrote this for convenience when developing on it, the method of development should change, by either referencing the source in the project instead of dist, or you have a separate watcher that watches the source files used in a build and runs `build`.

web/tests/fieldsets-parser-spec.js
wytrych: Yes I can load, but I can as well mock it

web/Gruntfile.js
wytrych: It happened to me a few times that I pushed and then noticed that I didn't run `build` before, so had to push again, I think it's fine to have a build task running on a `watch` can't see any harm

wytrych: Also I need to `build` because I browserify so I need that file for testing anyway

henrahmagix: Sure; I think there should be a separate watcher, so html changes runs `ngtemplates`, then you have a watcher on all the source files to run `build`.

wytrych:
@henrahmagix updated

henrahmagix:
The formly code I'm not fully aware of, so 👍  to that, and I have no other comments to make but would like another review to sanity check

henrahmagix:
Or, y'know, merge and fix up problems later =) This is still in progress so doesn't matter toooooooo much

henrahmagix:
👍

PR #27, opened 2016-08-19
Add form directive

wytrych:


wytrych:
@incuna/js Please review

web/app/scripts/src/directives/form.es.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.directive('surveysForm', [
+    FieldsetParser.componentName,
+    function (FieldsetParserService) {
+        return {
+            restrict: 'A',
+            scope: {
+                formStructure: '='
+            },
+            template: '<formly-form model="model" fields="fields"></formly-form>',
```
wytrych: For some strange reason this errors when used as an attribute. I filed a bug https://github.com/formly-js/angular-formly/issues/708

web/app/index.html
```diff
@@ -41,6 +41,10 @@
             <hr>
             <h5>Model:</h5>
             {{ form.model }}
+
+            <hr>
```
Minglee01: Does `<hr>` not have closing tags?

wytrych: Nope, it doesn't have any content

wytrych: https://developer.mozilla.org/en/docs/Web/HTML/Element/hr

wytrych:
@incuna/js bump

PR #29, opened 2016-08-22
Add URL for SurveyPostView

adam-thomas:
Because I forgot last week.

@incuna/backend review/merge please? :)

@wytrych fyi, I had to change the POST URL because I have two separate views handling GET and POST functionality

### DIFF
--- a/api-description/forms/pk/post.json
+++ b/api-description/forms/pk/post.json
@@ -1,6 +1,6 @@
 {
     "description": "Endpoint to answer a form",
-    "url": "http://localhost:8000/forms/1",
+    "url": "http://localhost:8000/forms/1/respond/",
     "fields": {
         "user_id": "User#20#",
         "user_responses": [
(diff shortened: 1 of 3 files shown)


api-description/forms/pk/post.json
```diff
@@ -1,6 +1,6 @@
 {
     "description": "Endpoint to answer a form",
-    "url": "http://localhost:8000/forms/1",
+    "url": "http://localhost:8000/forms/1/respond/",
```
KevinEtienne: I would have use `create` but let's see if frontend is happy.

KevinEtienne:
Waiting to see if we need to change https://github.com/incuna/incuna-surveys/pull/29#discussion_r75652927 but otherwise it looks good to merge.

api-description/forms/pk/post.json
adam-thomas: `/create` suggests you're making a new form, which you aren't here - you're submitting a user's response to it.

wytrych: Yeah `respond` is ok

adam-thomas:
Thanks :)

PR #30, opened 2016-08-22
Add an endpoint to retrieve a user's survey answers

adam-thomas:
Returns a dictionary of the user's latest answers to each fieldset in the survey.

@wytrych new (somewhat minimal) API description for your perusal :)


wytrych:
Looks ok

adam-thomas:
@incuna/backend bump?

surveys/urls.py
```diff
@@ -10,4 +10,9 @@
         views_api.SurveyPostView.as_view(),
         name='survey-post',
     ),
+    url(
+        r'^forms/(?P<pk>\d+)/respond/(?P<user_id>\w+)?$',
```
KevinEtienne: Do we know if a `user_id` can contain other character than a-z, can it contain underscores or dashes?

KevinEtienne: It looks like it can contains hashes from the api description.

surveys/views_api.py
```diff
@@ -23,3 +24,17 @@ def get_serializer(self, *args, **kwargs):
...
+    queryset = UserResponse.objects.all()
+
+    def dispatch(self, request, *args, **kwargs):
+        survey_pk = kwargs.pop('pk')
+        self.user_id = kwargs.pop('user_id')
+        self.survey = get_object_or_404(Survey.objects.all(), pk=survey_pk)
+        return super().dispatch(request, *args, **kwargs)
+
+    def list(self, request, *args, **kwargs):
+        data = self.get_queryset().latest_for_user(self.survey, self.user_id)
```
KevinEtienne: As `self.user_id` is coming for the url, it can contain a non existing `user_id` which would create an IndexError.

surveys/urls.py
adam-thomas: The `w` at least seems to accept underscores (going by the URL test), so I figure it should be fine.

surveys/views_api.py
adam-thomas: oh, good call, I should catch that and 404

surveys/urls.py
KevinEtienne: Thanks, let's wait until we have those user ids.

adam-thomas: It could be completely arbitrary, we should support as many characters as possible I think (do I need to do `.*` or whatever the correct syntax is?)

surveys/views_api.py
adam-thomas: Actually it's fine!  It just returns {} because `latest_for_user` uses a `defaultdict`.  Swanky.  I'll add a test for that.

surveys/urls.py
adam-thomas: Ahh, I changed it to `.+` which largely works, but some of the special characters get url-encoded:

```
    self.assertEqual(reversed_url, expected_url)
AssertionError: '/forms/42/respond/User%2320@!_%20Id' != '/forms/42/respond/User#20@!_ Id'
```

Do I need to un-encode the URLs in the view when I retrieve them?

adam-thomas:
Thanks :D

PR #32, opened 2016-08-24
Add API.post service.

maxpeterson:
For https://incuna-pm.atlassian.net/browse/OP-88


maxpeterson:
@incuna/frontend please review / merge

web/app/scripts/src/services/api.es.js
```diff
@@ -33,6 +33,12 @@ module.service(moduleProperties.componentName, [
             getForm: function (url) {
                 return $http.get(url)
                     .then((response) => response.data);
+            },
+            post: function (baseUrl, data) {
+                const endpoint = 'respond';
```
henrahmagix: Can you allow setting of `endpoint` in the configuration as a provider?

wytrych: I don't think it's a better solution than this, it's still out of sync with backend. A better solution would be to pass in the endpoints from the backend somehow.

maxpeterson: We have `const endpoint = 'forms';` just above. Seems a bit pointless to do this differently.

henrahmagix: They should all be able to be set via a provider =)

wytrych: I don't think it's such a great solution. Manually setting endpoints is not great, either way. The real solution would be to get the list of endpoints from the backend. But we can fix that later.

maxpeterson:
Included in #34

PR #33, opened 2016-08-25
Add date_created to the output of the latest_blah() methods

adam-thomas:
This is to support a display of when a user's survey was most recently updated in an export.


adam-thomas:
Thanks :)

PR #37, opened 2016-08-30
Add SurveyListView

KevinEtienne:
Add a `SurveyListView` which is a list of `SureyView` (detail). We might want to add `pagination` but we don't have the use case yet.


KevinEtienne:
@adam-thomas can you review please?

adam-thomas:
LGTM

PR #39, opened 2016-09-03
Escape twig templates and unique ids (across fieldsets).

maxpeterson:


maxpeterson:
@incuna/js please review / merge (cc @wytrych )

PR #42, opened 2016-09-06
Add not-empty class to field wrappers for styling

pandalion:
@incuna/js please review. We need the `not-empty` class on the wrapper of input, because input can't have pseudo element, in order to style. I think this should be in the library because it will be useful for styling across projects.

@wytrych is this what you meant about changing the field-id to match the id?

### DIFF
--- a/web/templates/twig-source/incuna-surveys/fields/free-text.html
+++ b/web/templates/twig-source/incuna-surveys/fields/free-text.html
@@ -1,3 +1,3 @@
-<div drf-form-field="to.fieldOptions" class="text" field-id="options.key">
+<div drf-form-field="to.fieldOptions" class="text {% raw %}{{ model[to.fieldSetId][options.key] ? 'not-empty' : '' }}{% endraw %}" field-id="{% raw %}{{ to.autoId }}{% endraw %}">
     <input type="text" class="text-input" id="{% raw %}{{ to.autoId }}{% endraw %}" ng-model="model[to.fieldSetId][options.key]" ng-required="to.fieldOptions.required">
 </div>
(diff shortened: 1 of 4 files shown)


web/templates/twig-source/incuna-surveys/fields/free-text.html
```diff
@@ -1,3 +1,3 @@
-<div drf-form-field="to.fieldOptions" class="text" field-id="options.key">
+<div drf-form-field="to.fieldOptions" class="text {% raw %}{{ model[to.fieldSetId][options.key] ? 'not-empty' : '' }}{% endraw %}" field-id="{% raw %}{{ to.autoId }}{% endraw %}">
```
grahamgilchrist: might be clearer to wrap the whole line in a `{%raw%}` rather than repeat it twice?

grahamgilchrist: oh wait they are different attributes. hmm maybe not then

wytrych: `field-id` shoudl be just `field-id="to.autoId"`

web/templates/twig-source/incuna-surveys/fields/number.html
```diff
@@ -1,3 +1,3 @@
-<div drf-form-field="to.fieldOptions" class="number" field-id="options.key">
+<div drf-form-field="to.fieldOptions" class="number {% raw %}{{ model[to.fieldSetId][options.key] ? 'not-empty' : '' }}{% endraw %}" field-id="{% raw %}{{ to.autoId }}{% endraw %}">
```
wytrych: Fix `field-id` here as above

web/templates/twig-source/incuna-surveys/fields/free-text.html
grahamgilchrist: you might also be able to do `ng-class="{'not-empty': model[to.fieldSetId][options.key]}"`

pandalion:
@wytrych amended

web/templates/twig-source/incuna-surveys/fields/free-text.html
grahamgilchrist: oh you can't `ng-class` because the directive is already doing it. I dislike directives which use `replace` for this reason!

grahamgilchrist:
:+1 when travis happy

pandalion:
thanks!!

PR #47, opened 2016-09-08
Percentage done add

Minglee01:
@incuna/js 

please review
for https://incuna-pm.atlassian.net/browse/OP-160

but dont merge ill add the tests after the code is ok


web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
+                }
+
+                const calculatePercentageComplete = function (completedQuestions) {
+                    let result = ((completedQuestions / scope.totalQuestionCount) * 100)
+                    if (!isNaN(result)) {
+                        scope.percentageComplete = Math.round(result) + '%';
+                    }
+                }
+
+                // Watching the model for changes.
```
wytrych: This comment is redundant

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
+                // Using true to compare the subelements.
+                scope.$watch('model', (answers) => {
+                    let completedQuestions = checkAnswered(answers);
```
wytrych: use `const`

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
+                                answered = answered + 1;
+                            }
+                        });
+                    });
+
+                    return answered;
+                }
+
+                const calculatePercentageComplete = function (completedQuestions) {
+                    let result = ((completedQuestions / scope.totalQuestionCount) * 100)
```
wytrych: use `const`

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
                             }
                         });
                     }
                 });
 
+                // Check and count the number of questions answered.
+                const checkAnswered = function (answers) {
+                    let answered = 0;
+                    angular.forEach(answers, function (answerGroup) {
+                        angular.forEach(answerGroup, function (item) {
```
wytrych: use native `forEach` instead of `angular.forEach`

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
+                            if (item > 0 || item.length > 0) {
+                                answered = answered + 1;
```
wytrych: `answered++;`

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
                             if (Object.keys(scope.model).length === 0) {
                                 scope.model = FieldsetParser.parseFormToModel(structure);
+                                countQuestionsTotal();
                             }
                         });
                     }
                 });
 
+                // Check and count the number of questions answered.
+                const checkAnswered = function (answers) {
```
wytrych: rename to something like `countNumberOfAnsweredQuestions` and remove comment as I think the name is self descriptive then

web/app/scripts/src/directives/form.es.js
wytrych: `numberOfCompletedQuestions` maybe?

web/app/scripts/src/directives/form.es.js
```diff
@@ -30,6 +30,17 @@ module.directive('surveysForm', [
             templateUrl: 'templates/incuna-surveys/form/survey-form.html',
             link: function (scope) {
                 scope.model = {};
+                scope.percentageComplete = 0 + '%';
+
+                const countQuestionsTotal = function () {
+                    const questions = scope.form.fieldsets;
+                    scope.totalQuestionCount = 0;
+
+                    angular.forEach(questions, function (question) {
```
wytrych: use native `forEach`

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
+                            }
+                        });
+                    });
+
+                    return answered;
+                }
+
+                const calculatePercentageComplete = function (completedQuestions) {
+                    let result = ((completedQuestions / scope.totalQuestionCount) * 100)
+                    if (!isNaN(result)) {
```
wytrych: I don't see a case where it would be `NaN`

web/app/scripts/src/directives/form.es.js
Minglee01: Dunno why you dont like comments i like them

wytrych: Comments usually try to make up for code that is not clear. In this case the comment can be easily removed by giving a better name to the function.

web/app/scripts/src/directives/form.es.js
wytrych: Ok if the are no questions, but in that case I would set it to `100%`

web/app/scripts/src/directives/form.es.js
Minglee01: even if its clear code i like to put comments if someone who was new to javascript the comments will help make it easier to read

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,40 @@ module.directive('surveysForm', [
...
+                    angular.forEach(answers, function (answerGroup) {
+                        angular.forEach(answerGroup, function (item) {
+                            if (item > 0 || item.length > 0) {
+                                answered = answered + 1;
+                            }
+                        });
+                    });
+
+                    return answered;
+                }
```
wytrych: missing semicolon

web/app/scripts/src/directives/form.es.js
wytrych: Sometimes comments can have value, but here the name of the function can describe exactly the same thing you are putting in the comment.

wytrych: http://butunclebob.com/ArticleS.TimOttinger.ApologizeIncode

Minglee01: only if the name of the function makes sense to everyone. what makes sense to you may not make sense to someone else

wytrych: What doesn't make sense in `countNumberOfAnsweredQuestions` ?

wytrych: You can come up with a better name though

web/app/scripts/src/directives/form.es.js
Minglee01: Whats the advantage of a native for each?

wytrych: We seem to be using to now everywhere, so for consistency. Other than I think it just reads cleaner:
`questions.forEach` instead of `angular.forEach(question,...)`

Minglee01:
@wytrych  think im done but i might need help to write the test for the functions.

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,49 @@ module.directive('surveysForm', [
...
+                //     }
+                // }
+                // answered is a number of type number
+                const countNumberOfAnsweredQuestions = function (answers) {
+                    let answered = 0;
+
+                    for (let groupKey in answers) {
+                        let answerGroup = answers[groupKey]
+                        for (let answerKey in answerGroup) {
+                            let answer = answerGroup[answerKey]
```
wytrych: This can be `const` and missing semicolon. Same comment for two lines above

web/app/scripts/src/directives/form.es.js
```diff
@@ -39,11 +50,49 @@ module.directive('surveysForm', [
...
+                    }
+
+                    return answered;
+                };
+
+                const calculatePercentageComplete = function (completedQuestions) {
+                    const result = ((completedQuestions / totalQuestionCount) * 100)
+                    if (!isNaN(result)) {
+                        scope.percentageComplete = Math.round(result) + '%';
+                    }
```
wytrych: the only case where `result` would be `NaN` is whern `totalQuestionsCount` is `0`. So what I think is the better solution is to make `scope.percentageComplete = '100%`;` in that case:
```
if (totalQuestoinsCount === 0) {
    scope.percentageComplete = '100%';
} else {
 const result = ....;
...
}
```
```

Comment (Minglee01) [inline web/app/scripts/src/directives/form.es.js]: well not really id prefer 0%.
Making it 100% may be misleading,

**Comment by wytrych** [comment]
Probably to test this you would need to 
1. Check that if the model is coming back the calculation is correct
2. Check that if there are no questions the result is 100%
3. Check that if when you update the questions the number changes

File: web/app/scripts/src/directives/form.es.js [hunk 12]
```diff
@@ -39,11 +50,49 @@ module.directive('surveysForm', [
...
+                // }
+                // answered is a number of type number
+                const countNumberOfAnsweredQuestions = function (answers) {
+                    let answered = 0;
+
+                    for (let groupKey in answers) {
+                        let answerGroup = answers[groupKey]
+                        for (let answerKey in answerGroup) {
+                            let answer = answerGroup[answerKey]
+                            if (answer > 0 || answer.length > 0) {
```
wytrych: What is this `asnwer > 0` check for? Also what happens when the answer is a string, is that possible?

web/app/scripts/src/directives/form.es.js
wytrych: To not make it flash it should be hidden with `ng-show`. I think if there aren't any questions then the result should be `100%` because you can't answer anything? If ti would be `0%` that would mean the user needs to answer something but nothing is there so that would be confusing imo. But maybe ask someone else

web/app/scripts/src/directives/form.es.js
```diff
@@ -30,6 +30,17 @@ module.directive('surveysForm', [
             templateUrl: 'templates/incuna-surveys/form/survey-form.html',
             link: function (scope) {
                 scope.model = {};
+                scope.percentageComplete = 0 + '%';
+                let totalQuestionCount = 0;
+
+                const countQuestionsTotal = function () {
+                    const questions = scope.form.fieldsets;
+
+                    questions.forEach((question) => {
```
maxpeterson: Could this use [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)  `questions.reduce((value, questions) => {value + question.fields.length} , 0)`

web/app/scripts/src/directives/form.es.js
Minglee01: if there arent questions this shouldnt show at all

wytrych: Ok, could you change the `isNan` condition to `if (totalQuestoinsCount !== 0) {` ? I think it''s more clear to understand

Minglee01:
@wytrych  @grahamgilchrist  please review.

grahamgilchrist:
@Minglee01 All looks good. Am happy to merge. 

One suggestion on the test, is maybe have tests to check the two extremes as well, e.g. that it correctly returns 0% and 100% when no questions and all questions answered. I find it useful with tests to cover the two extremes as well as a value in between, as you are more likely to prevent errors if someone refactors.

web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -0,0 +1,65 @@
...
+const module = angular.module(moduleProperties.moduleName, []);
+
+module.service(moduleProperties.componentName, [
+    function () {
+        this.countQuestionsTotal = function (form) {
+            const questions = form;
+            let totalQuestionCount = 0;
+
+            angular.forEach(questions, function (question) {
+                totalQuestionCount = totalQuestionCount + question.fieldGroup.length;
```
wytrych: `+=` ?

web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -0,0 +1,65 @@
...
+    moduleName: 'incuna-surveys.calculate-completion-percent',
+    componentName: 'calculateCompletionPercent'
+};
+
+const module = angular.module(moduleProperties.moduleName, []);
+
+module.service(moduleProperties.componentName, [
+    function () {
+        this.countQuestionsTotal = function (form) {
+            const questions = form;
```
wytrych: This seems redundant, can you just change the name of the argument instead?

web/tests/calculate-percentage-directive-spec.js
```diff
@@ -0,0 +1,55 @@
...
+        })
+        this.tpl = '<div calculate-percentage question-set="fields" model="getResponse"></div>';
+        this.compileDirective = function (tpl) {
+            this.elm = this.$compile(tpl)(this.scope);
+            this.scope.$digest();
+        };
+    });
+
+    it('should call this.calculateCompletionPercent.countQuestionsTotal with the questionSet', function() {
+        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
```
wytrych: These test do no test anything. You need an `expect` here. `spyOn` should be put inside `beforeEach` and here you need to do `expect(this.calculateCompletionPercent.countQuestionsTotal).toHaveBeenCalledWith(...)`

web/tests/calculate-percentage-survey-spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+    describe('countNumberOfAnsweredQuestions function', function () {
+        it('should return the total number of questions answered', function () {
+            expect(this.totalQuestionsAnswered).toBe(2);
+        });
+    });
+
+    describe('calculatePercentageComplete function', function () {
+        it('should return the 0 percentage when none answered', function () {
+            this.percentage = this.Parser.calculatePercentageComplete(0, this.totalQuestions);
```
wytrych: You don;t need `this` here as these are local variavbles. Can be just `const percentage`

wytrych:
@Minglee01 Can you at least fix the tests for the directive in a new PR? They currently don't test anything

PR #49, opened 2016-09-08
add missing hyphen in radio template so they actually work

hippogriffic:


hippogriffic:
uh shit am I meant to commit compiled templates? @wytrych

wytrych:
Hmm yeah probably although we're not using them, but you could

PR #52, opened 2016-09-08
add classes to radio template for styling

hippogriffic:
stuff stuff stuff


wytrych:
Stuff in `dist` is commited but here you have some strange files. Can you make a PR to remove them please?

PR #53, opened 2016-09-08
Handle missing response data.

maxpeterson:
This is to cope with changes to the survey form resulting in response data
that does not match the form.


maxpeterson:
@incuna/backend please review / merge

PR #55, opened 2016-09-09
Fix calc percent directive tests

Minglee01:
@wytrych  please help. functions arent being called.

### DIFF
--- a/web/tests/calculate-percentage-directive-spec.js
+++ b/web/tests/calculate-percentage-directive-spec.js
@@ -43,13 +43,21 @@ describe('calculatePercentage directive', function() {
             this.elm = this.$compile(tpl)(this.scope);
             this.scope.$digest();
         };
+        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
+        spyOn(this.calculateCompletionPercent, 'countNumberOfAnsweredQuestions').and.returnValue(answers);
     });
 
     it('should call this.calculateCompletionPercent.countQuestionsTotal with the questionSet', function() {
-        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
+        this.scope.field = questionSet;
+        this.scope.getResponse = getResponse;
+        this.compileDirective(this.tpl);
+        expect(this.calculateCompletionPercent.countQuestionsTotal).toHaveBeenCalled()
     });
 
     it('should call this.calculateCompletionPercent.countNumberOfAnsweredQuestions with the answer set', function() {
-        spyOn(this.calculateCompletionPercent, 'countNumberOfAnsweredQuestions').and.returnValue(answers);
+        this.scope.field = questionSet;
+        this.scope.getResponse = getResponse;
+        this.compileDirective(this.tpl);
+        expect(this.calculateCompletionPercent.countNumberOfAnsweredQuestions).toHaveBeenCalled()
     });
 });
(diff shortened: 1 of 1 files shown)


web/tests/calculate-percentage-directive-spec.js
```diff
@@ -43,13 +43,15 @@ describe('calculatePercentage directive', function() {
             this.elm = this.$compile(tpl)(this.scope);
             this.scope.$digest();
         };
+        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
+        spyOn(this.calculateCompletionPercent, 'countNumberOfAnsweredQuestions').and.returnValue(answers);
     });
 
     it('should call this.calculateCompletionPercent.countQuestionsTotal with the questionSet', function() {
-        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
```
wytrych: You need to call `this.compileDirective(this.template)` inside the test. You probaboy need to set  `this.scope.getResponse` and `this.scope.field`to something before compiling because otherwise the `$watch` won't trigger as the `model` and `question-set` will be empty

grahamgilchrist:
Looks ok to me, but maybe good if @wytrych checks also

Minglee01:
@wytrych is OK?

web/tests/calculate-percentage-directive-spec.js
```diff
@@ -43,13 +43,21 @@ describe('calculatePercentage directive', function() {
...
+        this.compileDirective(this.tpl);
+        expect(this.calculateCompletionPercent.countQuestionsTotal).toHaveBeenCalled()
     });
 
     it('should call this.calculateCompletionPercent.countNumberOfAnsweredQuestions with the answer set', function() {
-        spyOn(this.calculateCompletionPercent, 'countNumberOfAnsweredQuestions').and.returnValue(answers);
+        this.scope.field = questionSet;
+        this.scope.getResponse = getResponse;
+        this.compileDirective(this.tpl);
+        expect(this.calculateCompletionPercent.countNumberOfAnsweredQuestions).toHaveBeenCalled()
```
wytrych: What I would only change is that you can check that it has been called with the describe values. Use `toHaveBeenCalledWith()`

Minglee01: might need some help with that i tried earlier i was getting undefined.

Minglee01:
```
Firefox 48.0.0 (Mac OS X 10.10.0) calculatePercentage directive should call this.calculateCompletionPercent.countQuestionsTotal with the questionSet FAILED
	Expected spy countQuestionsTotal to have been called with [ Object({ 1: Object({ fieldGroup: [ 1, 2, 3 ] }), 2: Object({ fieldGroup: [ 1, 2, 3 ] }) }) ] but actual calls were [ undefined ].
```

web/tests/calculate-percentage-directive-spec.js
```diff
@@ -43,13 +43,21 @@ describe('calculatePercentage directive', function() {
             this.elm = this.$compile(tpl)(this.scope);
             this.scope.$digest();
         };
+        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
+        spyOn(this.calculateCompletionPercent, 'countNumberOfAnsweredQuestions').and.returnValue(answers);
     });
 
     it('should call this.calculateCompletionPercent.countQuestionsTotal with the questionSet', function() {
-        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
+        this.scope.field = questionSet;
```
wytrych: Maybe the problem is that you have `field` here but in the template you have `fields` ?

Minglee01:
@wytrych  thanks tests updated.

wytrych:
HWTI

PR #59, opened 2016-09-15
Include fields with a value of zero in the percentage calculation

grahamgilchrist:
This PR adjusts the percentage complete calculation to include questions which have a legitimate answer of the number zero.

Another effect of this change, is that questions answered as a blank string are now counted when they weren't before.


grahamgilchrist:
@incuna/js Pls review/merge

wytrych:
Would be nice to add a test for that

wytrych:
Nice, HWTI

grahamgilchrist:
Good idea. I'm not sure if we should be counting empty strings, but at least with the test we make the behaviour explicit and known.

PR #61, opened 2016-09-22
Enable put and patch on the latest / create view.

maxpeterson:


maxpeterson:
@incuna/backend is this even a good idea (cc @adam-thomas )?

surveys/serializers.py
```diff
@@ -6,6 +7,16 @@
...
 
 
+def update_nested(d, u):
+    for k, v in u.items():
+        if isinstance(v, collections.Mapping):
+            r = update_nested(d.get(k, {}), v)
+            d[k] = r
+        else:
+            d[k] = u[k]
+    return d
```
adam-thomas: First thing I see in this PR is this method...

adam-thomas: Actually, this could use a docstring and some more descriptive variable names, that'd help.

surveys/serializers.py
```diff
@@ -44,12 +55,18 @@ def get_fields(self):
         Dynamically generate the fields from the survey fieldsets.
         Returns a dictionary of {field_name: field_instance}.
         """
+
```
adam-thomas: Don't think this newline should be here? :)

maxpeterson:
Closing this until we identify we actually need this behaviour

adam-thomas:
:+1:

PR #63, opened 2016-09-23
Dry out surveys API

grahamgilchrist:
We had `api.getForm()` and `api.get()` methods which do exactly the same thing. This PR removes `getForm`, and dries out the use of the two.


grahamgilchrist:
@incuna/js Pls review/merge

wytrych:
HWTI

henrahmagix:
👍

PR #64, opened 2016-09-25
Add integer list field

maxpeterson:
Add functionality for a field that accepts multiple integers to be used to specify proportions such as a group of percentages.


maxpeterson:
@incuna/backend / @incuna/js please review

web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -0,0 +1,55 @@
...
+        return {
+            restrict: 'A',
+            scope: {
+                model: '=',
+                options: '=proportionField'
+            },
+            templateUrl: 'templates/incuna-surveys/form/proportion-field.html',
+            link: function (scope) {
+                scope.fields = [];
+                scope.$watch('options', (options) => {
```
wytrych: Should the watch be deregistered after usage?

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -0,0 +1,61 @@
...
+        this.calculateTotal = function (values) {
+            return Object.keys(values).reduce((value, key) => {
+                    let current = 0;
+                    if (angular.isNumber(values[key])) {
+                       current = parseInt(values[key], 10)
+                    }
+                    return value + current
+                },
+                0
+            );
```
wytrych: This indentation is getting a bit crazy here, would you mind splitting this up a bit?

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -0,0 +1,61 @@
...
+            });
+        };
+
+        this.calculateTotal = function (values) {
+            return Object.keys(values).reduce((value, key) => {
+                    let current = 0;
+                    if (angular.isNumber(values[key])) {
+                       current = parseInt(values[key], 10)
+                    }
+                    return value + current
```
wytrych: Semicolon here and two up

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -0,0 +1,61 @@
...
+                        id: `${autoId}-${index}`
+                    }
+                );
+            });
+        };
+
+        this.calculateTotal = function (values) {
+            return Object.keys(values).reduce((value, key) => {
+                    let current = 0;
+                    if (angular.isNumber(values[key])) {
```
wytrych: Should we be not accepting number strings as well?

wytrych: Also this will pass if you have a `NaN` for any reason so your sum will be affected

web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -0,0 +1,55 @@
...
+                            options.choices,
+                            options.fieldOptions,
+                            options.autoId
+                        );
+                    }
+                });
+
+                scope.$watch('options.fieldOptions.errors', (errors) => {
+                    if (errors) {
+                        ProportionField.addErrors(scope.fields, errors);
```
wytrych: You probably need to clear the errors too somehow?

web/tests/proportion-field-service-spec.js
```diff
@@ -0,0 +1,123 @@
...
+            {label: 'Two', required: false, id: '1-1'},
+            {label: 'Three', required: false, id: '1-2'},
+            {label: 'Four', required: false, id: '1-3'}
+        ];
+
+        this.values = {
+            0: 10,
+            1: 20,
+            2: 40,
+            3: 30
```
wytrych: Is this any different then an Array?

web/app/scripts/src/directives/proportion-field.es.js
maxpeterson: `addErrors` replaces the `errors` attribute so  clears the errors if there are none - I will add a test

web/tests/proportion-field-service-spec.js
maxpeterson: No but the system provides an object. Long term I think we should move away from array based referencing as it is not very scalable.

surveys/serializer_field_generators.py
```diff
@@ -4,6 +4,23 @@
 from .generators import BaseFieldGenerator as BaseBaseFieldGenerator
 
 
+class MultipleIntegerSerializer(serializers.Serializer):
+    """Dynamically generate a serializer with IntegerField based on the choices"""
```
maxpeterson: I didn't call it `Field` because it does not extend `Field` and although `Fields` are serializers they add some extra functionality.

maxpeterson:
@wytrych updated

web/tests/proportion-field-service-spec.js
```diff
@@ -117,6 +117,18 @@ describe('proportion-field service', function() {
...
+        it('should clear old errors from fields', function() {
+            const errors = ["A valid integer is required."];
+            const fields = [
+                {label: 'One', required: false, id: '1-0', errors: undefined},
+                {label: 'Two', required: false, id: '1-1', errors: errors},
+                {label: 'Three', required: false, id: '1-2', errors: undefined},
+                {label: 'Four', required: false, id: '1-3', errors: errors}
+            ]
+            this.proportionField.addErrors(fields, []);
+            expect(this.fields).toEqual(this.fields);
```
wytrych: This assertion seems a bit pointless

web/tests/proportion-field-service-spec.js
```diff
@@ -117,6 +117,18 @@ describe('proportion-field service', function() {
...
         });
+
+        it('should clear old errors from fields', function() {
+            const errors = ["A valid integer is required."];
+            const fields = [
+                {label: 'One', required: false, id: '1-0', errors: undefined},
+                {label: 'Two', required: false, id: '1-1', errors: errors},
+                {label: 'Three', required: false, id: '1-2', errors: undefined},
+                {label: 'Four', required: false, id: '1-3', errors: errors}
+            ]
```
wytrych: semicolon

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -30,12 +30,15 @@ module.service(moduleProperties.componentName, [
...
-            return Object.keys(values).reduce((value, key) => {
-                let current = 0;
-                if (angular.isNumber(values[key])) {
-                    current = parseInt(values[key], 10)
+            return Object.keys(values).reduce((total, key) => {
+                if (!values[key]) {
+                    return total;
                 }
-                return value + current
+                let value = values[key] && parseInt(values[key], 10);
```
wytrych: Can be `const` also the `values[key]` check is redundant since if falsy would have returned

wytrych: This would be sufficient:
```
const value = parseInt(values[key], 10);
if (Number.isNan(value)) {
     return total;
}
return total + value;
```

web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -0,0 +1,57 @@
...
+                    if (values) {
+                        ProportionField.addPercentages(scope.fields, values, scope.total);
+                    }
+                }, true);
+            }
+        };
+    }
+]);
+
+export default moduleProperties;
```
henrahmagix: Could you do this at the point of definition for `moduleProperties`?

henrahmagix: Also no need to export an object since you can export individual things and `import *` to get an object.
```js
export const name = 'my-module';
export const foo = 'foo';

export default name;
```
```js
import * as myModule from 'my-module';
console.log(myModule.foo) // 'foo'
```

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -0,0 +1,62 @@
...
+                );
+            });
+        };
+
+        this.calculateTotal = function (values) {
+            return Object.keys(values).reduce((total, key) => {
+                if (!values[key]) {
+                    return total;
+                }
+                let value = values[key] && parseInt(values[key], 10);
```
henrahmagix: You don't need `values[key] &&` because the above returns if `!values[key]`

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -0,0 +1,62 @@
...
+                if (Number.isNaN(value)) {
+                    return total;
+                }
+                return total + value;
```
henrahmagix: Why not, like `const value` in `addPercentages` below:
```js
return (total + value) || total;
```

web/tests/proportion-field-service-spec.js
henrahmagix: But an array is an object with zero index for keys, with the advantage of `forEach`.

web/tests/proportion-field-service-spec.js
henrahmagix: The linter should be running on test files

web/tests/proportion-field-service-spec.js
henrahmagix: Do you mean `expect(this.fields).toEqual(fields)`?

web/app/scripts/src/services/proportion-field.es.js
henrahmagix: You can combine with `isFinite()`

web/app/scripts/src/directives/proportion-field.es.js
henrahmagix: That's a good point, but perhaps you would want the fields to rebuild after an options change. However, this will only occur if the options thing was replaced, not if the object was altered, so practically this should only run once when `scope.options === undefined` becomes `scope.options === boundObject`

wytrych: The options won't change, not during the lifecycle of the directive.

wytrych: But good practice to remove unnecessary watchers

web/app/scripts/src/services/proportion-field.es.js
wytrych: I commented the same thing, this review system seems to be rubbish

web/app/scripts/src/directives/proportion-field.es.js
henrahmagix: Would be nice to provide a wrapper for that, if the explicit action is to only do things with `options` when it's available and then never again.

maxpeterson:
@wytrych / @henrahmagix I have no idea what github has done with the reviewing of this PR but could you take another look now.

I have added linting to the tests in https://github.com/incuna/incuna-surveys/pull/66

web/app/scripts/src/directives/proportion-field.es.js
maxpeterson: Added the deregister

web/app/scripts/src/services/proportion-field.es.js
maxpeterson: Updated indentation

web/app/scripts/src/services/proportion-field.es.js
maxpeterson: Done

web/app/scripts/src/services/proportion-field.es.js
maxpeterson: Updated

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -0,0 +1,62 @@
...
+        this.calculateTotal = function (values) {
+            return Object.keys(values).reduce((total, key) => {
+                if (!values[key]) {
+                    return total;
+                }
+                const value = parseInt(values[key], 10);
+                if (Number.isNaN(value)) {
+                    return total;
+                }
+                return total + value;
```
wytrych: So this function can be simpler:
```js
const value = parseInt(values[key], 10);
if (Number.isNaN(value)) {
    return total;
}
return total + value;
```

web/tests/proportion-field-service-spec.js
maxpeterson: ... but the api sends an object.

henrahmagix: 👍

web/tests/proportion-field-service-spec.js
maxpeterson: Don added lin 'ting in https://github.com/incuna/incuna-surveys/pull/66

web/app/scripts/src/services/proportion-field.es.js
maxpeterson: done

web/tests/proportion-field-service-spec.js
henrahmagix: >Long term I think we should move away from array based referencing as it is not very scalable.

Would like to discuss this more at some point

web/tests/proportion-field-service-spec.js
```diff
@@ -0,0 +1,141 @@
+describe('proportion-field service', function() {
+    beforeEach(function () {
+
+        angular.mock.module('incuna-surveys.proportion-field');
+
+        inject(function(proportionField) {
+            this.proportionField = proportionField;
+        });
+
+        this.fields = [
```
wytrych: I would rename this to `expectedFields` just to be clear what it is

web/app/scripts/src/directives/proportion-field.es.js
maxpeterson: not sure about this comment however I have copied the pattern used throughout this library

web/app/scripts/src/services/proportion-field.es.js
maxpeterson: done

web/tests/proportion-field-service-spec.js
wytrych: But I see you use it in different context as well so maybe dw

web/app/scripts/src/services/proportion-field.es.js
maxpeterson: Not sure I follow.

henrahmagix:
👍

web/app/scripts/src/directives/proportion-field.es.js
henrahmagix: I like to see `export` at the point of definition, but I know you can't `export default` a definition, so I'd like `export default foo` to be right after `const foo`. Or
```js
export const foo = 'foo';
export default foo;
```

web/app/scripts/src/services/proportion-field.es.js
henrahmagix: Since `addPercentages` is using falseyness to determine `NaN`, why not also do that here instead of using the explicit `Number.isNaN()`?

So
```js
if (Number.isNaN(value)) {
    return total;
}
return total + value;
```
could become
```js
return (total + value) || total;
```
because `total + value`, if `NaN`, would be falsey and it would move on to `return total`.

web/tests/proportion-field-service-spec.js
maxpeterson: They are also used as the fields passed to `addPercentages` and `addErrors`

web/app/scripts/src/services/proportion-field.es.js
henrahmagix: Unless the case of `total + value === 0` shouldn't be ignored.

web/tests/proportion-field-service-spec.js
maxpeterson: Maybe it is confusing that the function is destructively modifying the `fields` object.

surveys/serializer_field_generators.py
```diff
@@ -4,6 +4,31 @@
 from .generators import BaseFieldGenerator as BaseBaseFieldGenerator
 
 
+class BlankIntegerField(serializers.IntegerField):
```
maxpeterson: @incuna/backend is there a better / different way to do this?

maxpeterson: I think we need the same / similar functionality for all field generators based on `IntegerFields`  as the frontend will send an empty string when there is no value.

maxpeterson:
@adam-thomas / @wytrych any further comments.

adam-thomas:
I don't seem to be able to reply to the comment you @'d me in, but I'm fine with it as is.  I don't think I quite understand what's going on but I trust that it works. :P

surveys/serializer_field_generators.py
KevinEtienne: You can maybe try to set `initial=None` to the field see http://www.django-rest-framework.org/api-guide/fields/#initial

KevinEtienne: No it doesn't work.

PR #68, opened 2016-09-28
set up template folder structure properly

hippogriffic:
we need base templates that are extended in the regular templates so we can overwrite the regular templates locally and extended the base templates to be able to use the blocks


pandalion:
This looks good and sensible to me. Nice 😄

PR #70, opened 2016-10-16
Add missing translate and conditional submit button.

maxpeterson:


maxpeterson:
@incuna/frontend please review / merge

web/templates/twig-source/incuna-surveys/form/base/survey-form.html
```diff
@@ -5,5 +5,5 @@ <h1 ng-bind="form.name" class="form-name"></h1>
         <h2 ng-bind="form.description" class="form-desc"></h2>
     </header>
     <formly-form class="form-body" model="model" fields="fields"></formly-form>
-    <button class="button-submit" type="submit">Submit</button>
+    <button ng-show="form" class="button-submit" type="submit" translate>Submit</button>
```
henrahmagix: What about the H1 and H2? They're not being hidden when their contents would be empty if `form` is falsey, since `ng-bind="form.name"` would fill the element with nothing but the element would still show.

PR #73, opened 2017-02-22
return list of user responses

todd-dembrey:
Create methods on DRF serilaizers should return the created instances as this allows the serializer to  store them at `serializer.instance`.

`searializer.save`
https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L214

`serializer.create`
https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L906

@incuna/backend

### DIFF
--- a/surveys/serializers.py
+++ b/surveys/serializers.py
@@ -61,19 +61,21 @@ class SurveyResponseSerializer(serializers.Serializer):
         We also need to fill in the survey, which isn't supplied explicitly for each of
         the individual user response objects.
         """
-
+        responses = list()
         survey = validated_data.pop('survey')
         user_id = validated_data.pop('user_id')
         for fieldset_pk, answers in validated_data.items():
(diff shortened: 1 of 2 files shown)


surveys/serializers.py
```diff
@@ -39,6 +39,11 @@ class Meta:
 
 
 class SurveyResponseSerializer(serializers.Serializer):
+    def to_representation(self, instance):
+        if isinstance(instance, dict):
+            return super(SurveyResponseSerializer, self).to_representation(instance)
+        return {response.survey_id: response.answers for response in instance}
```
KevinEtienne: Can the instance be `None`?

todd-dembrey: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L261

todd-dembrey: (thats a no)

PR #74, opened 2017-03-06
Improved proportion

wytrych:
Chaning the proportion field by adding a slider. This will need some styling.

@incuna/js @maxpeterson Please review


web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -34,10 +34,7 @@ module.service(moduleProperties.componentName, [
                 return 0;
             }
             return Object.keys(values).reduce((total, key) => {
-                const value = parseInt(values[key], 10);
-                if (Number.isNaN(value)) {
-                    return total;
-                }
+                const value = parseInt(values[key], 10) || 0;
```
todd-dembrey: :+1:

todd-dembrey: I'm going now, but i like that change. Will give a better review in morning

web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -46,13 +46,17 @@ module.directive('proportionField', [
...
+                    if (!newModel || !oldModel) {
                         scope.model = {};
+                        return;
                     }
-                    scope.total = ProportionField.calculateTotal(values);
-                    if (values) {
-                        ProportionField.addPercentages(scope.fields, values, scope.total);
+
+                    const newSum = ProportionField.calculateTotal(newModel)
+                    const changedIndex = ProportionField.getChangedFieldIndex(newModel, oldModel);
```
maxpeterson: This only needs to be calculated if `newSum > 100)` so could be within the `if`.

web/app/scripts/src/services/proportion-field.es.js
```diff
@@ -55,6 +52,16 @@ module.service(moduleProperties.componentName, [
             });
         };
 
+        this.getChangedFieldIndex = function (newModel, oldModel) {
+            const newValues = Object.values(newModel);
+            const oldValues = Object.values(oldModel);
+
+            return newValues.reduce((sum, element, index) => {
```
maxpeterson: Can we use `Array.findIndex`: 

```newValues.findIndex((element, index) => element !== oldValues[index])```

 - it will return `-1` if all elements are the same.

Comment (maxpeterson) [inline web/app/scripts/src/services/proportion-field.es.js]: - are we using a polyfill?

File: web/templates/twig-source/incuna-surveys/form/base/proportion-field.html [hunk 4]
```diff
@@ -20,11 +12,15 @@
                         >
                     {% block input %}
                         <div class="proportion-input" integer-field model="model[$index]" id="field.id" form="form"></div>
+                        <div
+                            aif-slider-input
+                            model="model[$index]"
+                            ceiling="100"
```
maxpeterson: Could we calculate and expose the `remaining` and use it to limit the `ceiling` and `width` of the slider? 

```
    ceiling="{{ model[$index] + remaining }}"
    width="{{ (model[$index] + remaining)/100 }}%"
````


image

Comment (wytrych) [inline web/templates/twig-source/incuna-surveys/form/base/proportion-field.html]: That was my first approach, but the slider doesn't accept a width. Limiting the value gives us exactly the same result though. I will need to expose the width for css purposes however

File: web/app/scripts/src/services/proportion-field.es.js (hunk 3 above)
Comment (wytrych) [inline web/app/scripts/src/services/proportion-field.es.js]: We are using `babel-polyfill`, nice suggesion will change

File: web/templates/twig-source/incuna-surveys/form/base/proportion-field.html (hunk 4 above)
Comment (maxpeterson) [inline web/templates/twig-source/incuna-surveys/form/base/proportion-field.html]: I think the `width` would need to just be standard `css` asn would probably want to be on a containing div - I think someone from @incuna/frontend could look ta it but they would need the `remaining` value in the scope - possibly using a simple directive to set the `model[$index] + remaining` value.

File: web/app/scripts/src/services/proportion-field.es.js (hunk 3 above)
Comment (maxpeterson) [inline web/app/scripts/src/services/proportion-field.es.js]: Using `reduce` is a neat solution to the problem when `findIndex` is not available.

File: web/templates/twig-source/incuna-surveys/form/base/proportion-field.html (hunk 4 above)
Comment (wytrych) [inline web/templates/twig-source/incuna-surveys/form/base/proportion-field.html]: So currently the allocated amount (that's what's needed to set the with of the div) is available in the proportion field template with: `total - model[$index]`.

**Comment by wytrych** [comment]
Improved

File: web/tests/proportion-field-directive-spec.js [hunk 5]
```diff
@@ -97,33 +97,34 @@ describe('proportion-field directive', function () {
...
-            spyOn(this.proportionField, 'calculateTotal').and.returnValue(this.total);
-            spyOn(this.proportionField, 'addPercentages');
             this.scope.options = this.options;
             this.scope.model = this.model;
             this.compileDirective();
             this.isolated = this.elm.isolateScope()
             this.scope.$digest();
         });
 
-        it('should call proportionField.calculateTotal and set scope.total', function () {
```
maxpeterson: Should we restore this test?

web/tests/proportion-field-service-spec.js
```diff
@@ -105,27 +105,6 @@ describe('proportion-field service', function () {
 
     });
 
-    describe('addPercentages', function () {
-        it('should add percentages to the fields based on the values and total', function () {
-            this.proportionField.addPercentages(this.fields, this.values, 100);
```
maxpeterson: Should we remove [proportionField.addPercentages](https://github.com/incuna/incuna-surveys/blob/c82a99375bcdc2634f3b491d261778bd71d87d74/web/app/scripts/src/services/proportion-field.es.js#L45) if it is no longer used or tested.

maxpeterson:
Are the tests being run correctly? Travis says "Executed 0 of 70 SUCCESS" - https://travis-ci.org/incuna/incuna-surveys/jobs/208532462 which seems odd

web/tests/proportion-field-directive-spec.js
```diff
@@ -96,34 +96,44 @@ describe('proportion-field directive', function () {
...
         });
 
-        it('should call addPercentages', function () {
-            expect(this.proportionField.addPercentages).toHaveBeenCalledWith(
-                this.isolated.fields,
-                this.model,
-                this.total
-            )
+        it('should call proportionField.calculateTotal and set scope.total', function () {
+            spyOn(this.proportionField, 'calculateTotal').and.returnValue(this.total);
```
maxpeterson: I can't see where this.total gets set - Possibly it is `undefiend` maybe it should be a variable local to this function?

maxpeterson:
It looks like Travis is running tests but seem a bit odd!

wytrych:
I changed the reporting style, should look less confusing now

maxpeterson:
Thanks @wytrych , sorry for all the small tweaks

wytrych:
Thanks!

PR #75, opened 2017-03-09
Fix slider by using options

wytrych:
The rzSlider was using a throttle so it didn't update the model values fast enough. This solution limits the slider via it's options.


wytrych:
@incuna/js Please review

web/templates/generated/incuna-surveys/form/base/proportion-field.html
```diff
@@ -11,16 +11,19 @@ <h4 class="title" ng-bind="title"></h4>
...
+                                clearable-rz-slider
+                                model="model[$index]"
+                                rz-slider-options="optionsPerSlider[$index]"
+                                slider-low-label="0%"
+                                slider-high-label="100%"
+                                ceiling="100"
+                            ></div>
+
+                            <span class="allocated-bar" style="width: {{ total - model[$index] }}%;"></span>
+                        </div>
```
Minglee01: can u remove one of the blank lines ?

Minglee01: nvm

web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -53,14 +66,19 @@ module.directive('proportionField', [
...
+                    scope.total = newSum;
 
                     if (newSum > 100) {
                         scope.total = 100
                         const changedIndex = ProportionField.getChangedFieldIndex(newModel, oldModel);
                         scope.model[changedIndex] -= newSum - 100;
                     }
+
+                    scope.fields.forEach((val, i) => {
+                        scope.optionsPerSlider[i].maxLimit = 100 - newSum + scope.model[i];
```
maxpeterson: What if `newSum > 100`? Don't we need to set  `newSum` to 100 in this case?

web/templates/generated/incuna-surveys/form/base/proportion-field.html
```diff
@@ -11,16 +11,19 @@ <h4 class="title" ng-bind="title"></h4>
...
-                        ></div>
-                    
+                        <div class="proportion-slider">
+                            <div
+                                clearable-rz-slider
+                                model="model[$index]"
+                                rz-slider-options="optionsPerSlider[$index]"
+                                slider-low-label="0%"
+                                slider-high-label="100%"
+                                ceiling="100"
```
maxpeterson: Should `ceiling` be specified here, as well as the `ceil` in `optionsPerSlider`?

web/app/scripts/src/directives/proportion-field.es.js
wytrych: Hmm good point

maxpeterson: Will it cause another (watch) iteration? In which case would it be corrected.

wytrych: Inside the previous `if` the model is being changed on the altered field and that will trigger a `watch`, so it will be rerun with a correct sum.

web/templates/generated/incuna-surveys/form/base/proportion-field.html
wytrych: I think the
```
slider-low-label="0%"
slider-high-label="100%"
ceiling="100"
```
bits are redundant

PR #76, opened 2017-03-20
Prevent calculate percentage directive setting percentageComplete to infinity.

Minglee01:
@todd-dembrey 
@maxpeterson 

please review
thanks


todd-dembrey:
Would it have been better to ensure that the calculation can handle zero values? You are separating the logic otherwise.

Also would make it easy to write a few tests :+1:

https://github.com/incuna/incuna-surveys/blob/1698b640f8f9d7ce2e4fa70b40246e0914cc61c3/web/app/scripts/src/services/calculate-percentage.es.js#L53

web/app/scripts/src/directives/calculate-percentage.es.js
```diff
@@ -29,10 +29,12 @@ module.directive('calculatePercentage', [
                 });
 
                 // Using true to compare the sub-elements.
-                scope.$watch('model', (answers) => {
-                    const numberOfCompletedQuestions = CalculateCompletion.countNumberOfAnsweredQuestions(answers);
-                    scope.percentageComplete = CalculateCompletion.calculatePercentageComplete(numberOfCompletedQuestions, totalQuestionCount);
-                }, true);
+                if (totalQuestionCount > 0) {
```
todd-dembrey: Should this be reverted now?

Minglee01: we dont want to trigger that watch. untill we have a question count

Minglee01: cause a watch makes the app slower

maxpeterson: The `link` is only run once (per directive on the page) so I would expect `totalQuestionCount` to always be `0` during the `link` function. I think you need to put the optimisation in the watch:
```
                scope.$watch('model', (answers) => {
                    if (totalQuestionCount == 0) {
                      return;
                    }
                    const numberOfCompletedQuestions = CalculateCompletion.countNumberOfAnsweredQuestions(answers);
                    scope.percentageComplete = CalculateCompletion.calculatePercentageComplete(numberOfCompletedQuestions, totalQuestionCount);
                }, true);
```

web/tests/calculate-percentage-directive-spec.js
```diff
@@ -45,20 +42,17 @@ describe('calculatePercentage directive', function () {
             this.elm = this.$compile(tpl)(this.scope);
             this.scope.$digest();
         };
-        spyOn(this.calculateCompletionPercent, 'countQuestionsTotal').and.returnValue(questionSet);
-        spyOn(this.calculateCompletionPercent, 'countNumberOfAnsweredQuestions').and.returnValue(answers);
     });
 
-    it('should call this.calculateCompletionPercent.countQuestionsTotal with the questionSet', function () {
```
maxpeterson: What has happened to this test?

web/app/scripts/src/directives/calculate-percentage.es.js
Minglee01: Do we want to run a watch if question count is 0?

Minglee01: Ok i get you

web/tests/calculate-percentage-directive-spec.js
Minglee01: I removed it as i didnt think it was that useful

Minglee01: The test i added implys the same thing. I can readd

Minglee01: The test i added implys the same thing. I can readd

maxpeterson:
THis good to merge if you are done making changes @Minglee01

PR #79, opened 2017-04-25
Change multi slider heading to span to match other heading style

brynley:
Needed for https://github.com/incuna/incuna-optimise/pull/414

@incuna/frontend Please can someone review 👾


hippogriffic:
I think you'll need to run grunt to compile the templates and commit those :)

hippogriffic:
looks like you've got an excited js test fail too :(

Minglee01:
Check the failed test might just b looking for a h4

wytrych:
It is because the test is looking for an `h4` so you need to update the test to look for `span.title` instead, or just `.title` to be more future proof

Minglee01:
Im gonna merge this. @wytrych  we tried `.title` i dont think this available at that point in the test

wytrych:
That's because `jqlite` `find()` works only for tag names, `expect(this.elm[0].querySelectorAll('.title').length).toEqual(1);` would work and I encourage you to bring the test back - it's always better to fix the test instead of removing it.

PR #81, opened 2017-04-28
Unrequire the answers field in the admin.

adam-thomas:
The answers field shouldn't be required for anything other than choice fields (radio/checkbox).


adam-thomas:
boimp

PR #82, opened 2017-06-08
Fix optimise null error

maxpeterson:


web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -66,6 +66,16 @@ module.directive('proportionField', [
                         return;
                     }
 
+                    if ( Object.values(newModel).every( (x) => { return x === null; } ) ) {
```
maxpeterson: You don't need the curleys `{}` or the return if there is only one command `Object.values(newModel).every((x) => x === null)`

web/app/scripts/src/directives/proportion-field.es.js
```diff
@@ -66,6 +66,16 @@ module.directive('proportionField', [
...
+                        return;
+                    }
+
+                    for ( var [k, v] of Object.entries(newModel) ) {
```
maxpeterson: We normally try to use `forEach`: `Object.entries(newModel).forEach((k, v) => {...})`

PR #85, opened 2017-08-07
Import questions calc

Minglee01:
@maxpeterson 

https://incuna-pm.atlassian.net/browse/OP-410


web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -15,12 +15,17 @@ const module = angular.module(moduleProperties.moduleName, []);
...
         this.countQuestionsTotal = function (form) {
-            const questions = form;
             let totalQuestionCount = 0;
-
-            angular.forEach(questions, function (question) {
-                totalQuestionCount = totalQuestionCount + question.fieldGroup.length;
+            angular.forEach(form, function (question) {
+                angular.forEach(question.fieldGroup, function (field) {
+                    if (field.templateOptions.fieldOptions.important === true) {
+                        totalQuestionCount = totalQuestionCount + 1;
```
maxpeterson: Could use `totalQuestionCount++` OR `totalQuestionCount += 1`

web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -15,12 +15,17 @@ const module = angular.module(moduleProperties.moduleName, []);
 
 module.service(moduleProperties.componentName, [
     function () {
+        let qKeys = [];
```
maxpeterson: It seems a bit risky to have `qKeys` defined like this, storing state on the service, outside the methods `countQuestionsTotal` and `countNumberOfAnsweredQuestions`

maxpeterson: I would replace `countQuestionsTotal` with a function that finds the important questions (e.g. what you have but define `qKeys` within the function and return `qKeys` from the function:

```
        this.ImportQuestionKeys = function (form) {
            let qKeys = [];
            angular.forEach(form, function (question) {
                angular.forEach(question.fieldGroup, function (field) {
                    if (field.templateOptions.fieldOptions.important === true) {
                        qKeys.push(field.key)
                    }
                });
            });

            return qKeys;
        };
```

You can then simply get the count with `countQuestions = ImportQuestionKeys(form).length`

And you can pass the `qKeys` returned by `ImportQuestionKeys` to  `countNumberOfAnsweredQuestions` to avoid storing the `qKeys` in the service.

web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -15,20 +15,22 @@ const module = angular.module(moduleProperties.moduleName, []);
...
                     }
                 });
             });
 
-            return totalQuestionCount;
+            return qKeys;
+        };
+
+        this.countQuestionsTotal = function (form) {
+            return ImportantQuestionKeys(form).length;
```
wytrych: You are missing `this` - `this.ImportantQuestionKeys`

web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -42,6 +44,8 @@ module.service(moduleProperties.componentName, [
         this.countNumberOfAnsweredQuestions = function (answers) {
             let answered = 0;
 
+            const qKeys = ImportantQuestionKeys(form)
```
wytrych: Same here, need `this`

Minglee01:
@maxpeterson

web/app/scripts/src/services/calculate-percentage.es.js
```diff
@@ -34,15 +37,21 @@ module.service(moduleProperties.componentName, [
...
+            }
+
             for (const groupKey in answers) {
                 const answerGroup = answers[groupKey];
                 for (const answerKey in answerGroup) {
                     const answer = answerGroup[answerKey];
-                    if (angular.isDefined(answer) && answer !== null) {
-                        answered++;
+                    if (qKeys.indexOf(parseInt(answerKey)) !== -1) {
+                        if (answer && answer !== null || answer === 0) {
```
maxpeterson: Where did the ` || answer === 0` come from? It doesn't look like it was there before?

maxpeterson: `parseInt` should really have a second argument `parseInt(answerKey, 10)`

Minglee01: `|| answer === 0` i added that. Previously it wasnt counting properly for me.
i was getting 100% without answering anything

web/tests/calculate-percentage-directive-spec.js
```diff
@@ -42,32 +92,33 @@ describe('calculatePercentage directive', function () {
...
-        expect(this.calculateCompletionPercent.countQuestionsTotal).toHaveBeenCalledWith(questionSet)
+        expect(this.calculateCompletionPercent.getImportantQuestionKeys).toHaveBeenCalledWith(questionSet)
     });
 
-    it('should return the total number of questions', function () {
-        const questionCount = this.calculateCompletionPercent.countQuestionsTotal(questionSet);
-        expect(questionCount).toEqual(6);
+    it('should return an array of keys', function () {
+        const keys = this.calculateCompletionPercent.getImportantQuestionKeys(questionSet);
+        expect(keys).toEqual([ 2, 3, 2, 3 ]);
```
maxpeterson: It is interesting that the keys could be repeated. Maybe it should use a [Set](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set)? - It would probably simplify the the service code to use `let qKeys = new Set();` ... `qKeys.add(field.key)` and `qKeys.has(answerKey)`

maxpeterson: It would be a minor efficiency improvement but would probably be more readable.

